PR #2, opened 2016-04-11
Improving CroneExpression.java

alexVengrovsk:
It changed, becouse in its previous state it worked wrong. Instead of returning an index of the next whitespace it returned a string length from the start position to the end of a string. I don't think, that it is ok.

### DIFF
--- a/quartz-core/src/main/java/org/quartz/CronExpression.java
+++ b/quartz-core/src/main/java/org/quartz/CronExpression.java
@@ -19,6 +19,7 @@ package org.quartz;
 
 import java.io.Serializable;
 import java.text.ParseException;
+import java.util.Collection;
 import java.util.Calendar;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


zemian:
Thanks for the PR @alexVengrovsk , but the changes won't compile. Can you review it again?

Also I like the improvement on eliminate duplicate getExpressionSet/ListSummary with single method. However I don't think the findNextWhiteSpace refactoring is worth it. On top of it, the built-in Character.isWhitespace is doing more than existing code, so I am not sure about this yet.

zemian:
@alexVengrovsk 

zemian:
Hi, I like to clean up the pending PR list. This PR has no updates for a long time and I do not think it's enough to add value. I will close this PR if no further update receive in few days.

zemian:
Close as invalid

Issue #6, opened 2016-05-02
Cannot create tables in MySQL with InnoDB and UTF8mb4 using `tables_mysql.sql`


krrg:
InnoDB has a maximum size of 767 bytes on a single column index. When using a `VARCHAR(200)` as one of the columns in a multicolumn primary key, this limit still is in force.

When the character encoding is set to `UTF8mb4`, the `VARCHAR(200)` data type required 800 bytes, which is greater than the allowed 767 by InnoDB.

This should be easy to fix by reducing the `VARCHAR(200)` columns to `VARCHAR(190)` columns, which will fit inside the 767 byte limit with any encoding that uses 4 bytes or less.

For limitations on InnoDB, see http://dev.mysql.com/doc/refman/5.6/en/innodb-restrictions.html

jhouserizer:
See #7

zemian:
PR merged.

PR #7, opened 2016-05-02
Fix for "Cannot create tables in MySQL with InnoDB and UTF8mb4"

krrg:
See https://github.com/quartz-scheduler/quartz/issues/6


disaacson:
I wouldn't mind if these fields were even smaller. 380 total characters for an identifier is pretty large. Should we reduce the size to 100 chars wide?

krrg:
I can't imagine a situation in which a 190 character identifier would be needed or would be valuable.

jhouserizer:

poojits:
`CALENDAR_NAME` should also be changed to`VARCHAR(190)` in `QRTZ_CALENDARS` and `QRTZ_CALENDARS` as there are indices on `CALENDAR_NAME` as well. Also, one should also modify `tables_mysql_innodb.sql` as well.

zemian:
Thanks for the PR @krrg 
Since these are InnoDB specific, can you modify your PR to tables_mysql_innodb.sql file instead? We can leave the original tables_mysql.sql if that does not bother anyone. Also can you please also update the ones mentioned by @poojits ?

krrg:
@jhouserizer  I agree to the grant statement. 

I'll go ahead and implement the changes listed by @poojits and move the changes into the `tables_mysql_innodb.sql` and update the PR as soon as I have a spare cycle (should be today).

zemian:
Looks good @krrg Thanks for the contribution!

PR #11, opened 2016-05-09
QTZ-483 - If JobListener throws an Exception, job is blocked indefinitely

cbgrilo:
This relates to the issue reported in https://jira.terracotta.org/jira/browse/QTZ-483.

I've found this to be a problem especially when using the DisallowConcurrentExecution annotation, since with it, as soon as any exception is throw by the Job Listener, the current job is left blocked and no other jobs are allowed to run afterwards.

I've fixed the issue for JobListeners in JobRunShell, however this might also occur when using TriggerListeners (line 248). Since I had no previous issues with them I did not test that specific case or made any further changes.

### DIFF
--- a/quartz-core/src/main/java/org/quartz/core/JobRunShell.java
+++ b/quartz-core/src/main/java/org/quartz/core/JobRunShell.java
@@ -171,6 +171,8 @@ public class JobRunShell extends SchedulerListenerSupport implements Runnable {
                 // notify job & trigger listeners...
                 try {
                     if (!notifyListenersBeginning(jec)) {
+                    	CompletedExecutionInstruction instCode = trigger.executionComplete(jec, null);
+                    	qs.notifyJobStoreJobComplete(trigger, jobDetail, instCode);
                         break;
                     }
                 } catch(VetoedException ve) {
@@ -222,6 +224,8 @@ public class JobRunShell extends SchedulerListenerSupport implements Runnable {
 
                 // notify all job listeners
                 if (!notifyJobListenersComplete(jec, jobExEx)) {
+                	CompletedExecutionInstruction instCode = trigger.executionComplete(jec, jobExEx);
+                	qs.notifyJobStoreJobComplete(trigger, jobDetail, instCode);
                     break;
                 }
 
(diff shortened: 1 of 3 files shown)


jhouserizer:

cbgrilo:
@jhouserizer I agree to the grant statement.

quartz-core/src/main/java/org/quartz/core/JobRunShell.java
```diff
@@ -171,6 +171,8 @@ public void run() {
                 // notify job & trigger listeners...
                 try {
                     if (!notifyListenersBeginning(jec)) {
+                    	CompletedExecutionInstruction instCode = trigger.executionComplete(jec, null);
+                    	qs.notifyJobStoreJobComplete(trigger, jobDetail, instCode);
```
zemian: Hm.. calling job complete when we haven't actually run the job?

jhouserizer:
There's certainly a corner case here - but I believe it needs more thought before just jumping in with this fix.  Indeed, as @zemian  points out, there are historical (expected behavior) considerations, also aside from that I'm not sure I believe the new semantics would meet the expectations of all future users either.

jhouserizer:
After some review and thought, I have these comments and points to add:

* The API contract/javadoc of the JobListener and TriggerListener interfaces don't state what behavior would be guaranteed in this case.  

* Depending upon the nature of the listener and the job, it may or may not be correct to do any of the following things when the listener throws a runtime exception:
-- put the job (its triggers) into ERROR state without even trying to execute it  (because it would be wrong to execute the job if the listeners couldn't succeed in their work)
-- act as if the job execution as vetoed by a trigger listener - don't execute the job, but don't put it (its triggers) in error state
-- ignore the exception from the listener and try executing the job (because the listener isn't doing anything vital to the job's execution)

If we look at examples of listeners that ship with quartz (e.g. in the plugins, etc.) and of those custom listeners that we've seen users have "in the field", then we know there are cases where a listener failure has critical ramifications.     We also have examples where listeners are doing mundane side things like logging or statistics.  So there is support for and against all of the above being the correct thing to do.

Setting aside those points for a bit, the problem this reported issue and patch are trying to resolve relates to the triggers of the job being left in the "BLOCKED" state.  Clearly that seems wrong and should be fixed.  However the patch "fixes" that by acting as if the job executed successfully, when it actually didn't execute at all, so clearly that's wrong.

zemian:
Hi, I like to clean up the pending PR list. This PR has no updates for a long time and I do not think it's enough to add value. I will close this PR if no further update receive in few days.

zemian:
Close as invalid

PR #15, opened 2016-05-24
TAB-4700 : Remove update-checker feature

rishabhmonga:


jhouserizer:
Looks good!  Thanks.

chrisdennis:
There are a lot of changes to unrelated lines in the examples properties files.  I think there all just the removal of commented lines, but the timing seems weird. Was this intentional?

rishabhmonga:
Had received a comment from Louis on the CR I had created earlier, suggesting that we could loose the commented lines in the properties files

chrisdennis:
Cool - thanks for following up.

ljacomet:
that's interesting :wink:

My other comment was mostly to not add commented lines about the update check since it is completely removed. I did not look at the context and did not realise there were already commented lines.

But hey, @jhouserizer found it fine!

Issue #19, opened 2016-05-31
License issue with c3p0


danielgratzl:
Quartz scheduler is released under the Apache 2.0 license but has a dependency on c3p0 (0.9.1.1) which uses LGPL. In my understanding this is not allowed http://www.apache.org/legal/resolved.html#category-x
Newer versions of c3p0 declare a dual license LGPL/EPL which should be be fine to statically link against without violating the Apache license model.

jhouserizer:
We're planning to move to newer c3p0 with the next release.   By the way, that list isn't strictly about license compatibility in general, bur rather what Apache organization advises it's own products to distribute (and Quartz is not an Apache product) - because of differences in terms - not that there's any problem legally with including them together, but that those consuming the product may not like the mixed terms.

danielgratzl:
Alright, thanks for the feedback. Even though i think LGPL would require you to not link the library statically against your code. But I actually do not know enough about this whole license thing, we're just trying to get our dependencies right :)
Thanks for the great work on quartz!

jhouserizer:
See #20

msteiger:
I stumbled over this a few days, because I was surprised to find c3p0 in my classpath despite the fact that I'm using DBCP2 as a connection pool.

I'm surprised to see a JDBC connection pool being explicitly required by a job scheduler. Would it make sense to externalize it into a plugin? Then, users interested in that feature could add `quartz-jdbc-pool` (if that would be its name) instead of `quartz-core` while other might just stick to the plain version and add the pool of their choice and in the version they need.

This is probably linked to #20 - there are always users who want to use a different pool. Having multiple ones floating around isn't ideal imho.

zemian:
@msteiger Our default out of box `org.quartz.utils.PoolingConnectionProvider` uses c3p0 inner class thus the hard dependency. But if you were to provide your own `ConnectionProvider`, then you may exclude the c3p0 from Maven dependency in your project like this:

```
        <dependency>
            <groupId>org.quartz-scheduler</groupId>
            <artifactId>quartz</artifactId>
            <version>${quartz.version}</version>
            <exclusions>
                <exclusion>
                    <groupId>c3p0</groupId>
                    <artifactId>c3p0</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

jhouserizer:
won't fix with quartz 2.2.x , c3p0 will be replaced with 2.3.x.

PR #20, opened 2016-06-01
Add support for HikariCP connection pool

lorban:


zemian:
Hi @lorban ,

I have looked into your PR and besides fixing the conflict, I have few concerns.

1) I am not sure introducing provider predefined names (using c3p0 or hikari) config property is a clean way to do this. We already have .class to provide a implementation, so it makes more sense just to let this classname value to drive the rest of config bean set. Why need another short name?

2) It seems that the old PoolingConnectionProvider (old c3p0) iml actually initialized the underneath DataSource in the constructor, and then leave the actual initialize() method empty. I think these should be reversed, and provide a empty constructor so that we can load the provider same way as as any custom provider.

3) Some one mentioned in an Issue #19 that we have hard (compile) dependency on c3p0 with quartz, so I am hoping this PR change can also remove that. We can continue to package a CP package jar if we need to, but the maven pom dependencies on "quartz" should not contains it for client. Just like a SLF4J, they have to pick an implementation for their need.

zemian:
Hi @lorban ,

I have looked into this PR and have few concerns (besides the file conflicts)

1) Adding new provider short name (eg: "c3p0" or "hiraki") is unnecessary and might be confusing to users. We already have the "connectionProvider.class." property, and this along should be unique and able to drive what's needed for rest of the bean properties.

2) The existing PoolingConnectionProvider (c3p0 impl) actually invoke initialization code in the constructor, and then leave the ConnectionProvider interface's "initialize()" method empty. This should really be reversed. We should provide a default empty constructor for c3p0 and use "initialize" properly. I would like to see our own provider impl should work load and work just like any other custom provider.

3) In Issue#19 someone points out we have a hard (compile) c3p0 maven dependency with quartz. I was hoping this PR can also address that. Would it make sense to not couple this dependency?

amilarajans:
+1

jhouserizer:
see #126 , which replaces this PR.  Thanks, @lorban .

PR #25, opened 2016-06-13
Reset trigger state from error state

marwtki:
When our triggers get into an error state, I want to reset the trigger state after fixing the underlying problem. I could modify the database directly but would rather go through an official API.


disaacson:
+1

jjudd:
Curious what the status of this is as it has been a few months

alangshall:
+1

jhouserizer:

jhouserizer:
Hi @marwtki,

Thanks!

ramaraochavali:
+1. Good Feature.

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
zemian: IDE generated files should not be checkin

quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
```diff
@@ -1598,6 +1598,37 @@ public TriggerState getTriggerState(Connection conn, TriggerKey key)
...
+     * </p>
+     *
+     * @see TriggerState#NORMAL
+     * @see TriggerState#PAUSED
+     * @see TriggerState#COMPLETE
+     * @see TriggerState#ERROR
+     * @see TriggerState#NONE
+     */
+    public void resetTriggerState(final TriggerKey triggerKey) throws JobPersistenceException {
+        executeWithoutLock(
```
zemian: I think we should use executeInLock instead.

quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
```diff
@@ -1598,6 +1598,37 @@ public TriggerState getTriggerState(Connection conn, TriggerKey key)
...
+                        resetTriggerState(conn, triggerKey);
+                        return null;
+                    }
+                });
+    }
+
+    public void resetTriggerState(Connection conn, TriggerKey key)
+            throws JobPersistenceException {
+        try {
+            getDelegate().updateTriggerState(conn, key, STATE_WAITING);
```
zemian: For audit sake, I think we should retrieve the old value and LOG it before reset new value.

quartz-core/src/main/java/org/quartz/simpl/RAMJobStore.java
```diff
@@ -665,6 +665,32 @@ public TriggerState getTriggerState(TriggerKey triggerKey) throws JobPersistence
...
+    public void resetTriggerState(TriggerKey triggerKey) throws JobPersistenceException {
+        synchronized (lock) {
+            TriggerWrapper tw = triggersByKey.get(triggerKey);
+
+            // does the trigger exist?
+            if (tw == null || tw.trigger == null) {
+                return;
+            }
+            tw.state = TriggerWrapper.STATE_WAITING;
+            applyMisfire(tw);
```
zemian: 1. Why are we calling applyMisfire here? I don't think resetTriggerState should imply misfire.

2. Also, for RAMJobStore, the initial state is set to null, so I think reset it to null instead of STATE_WAITING is more consistent. As long as it is added to the timeTriggers set is fine.

quartz-core/src/main/java/org/quartz/core/QuartzScheduler.java
```diff
@@ -1589,6 +1589,19 @@ public TriggerState getTriggerState(TriggerKey triggerKey) throws SchedulerExcep
 
     /**
      * <p>
+     * Reset the current state of the identified <code>{@link Trigger}</code>.
```
zemian: Should we clarify what value will the trigger state be after reset?

terracotta/bootstrap/src/main/java/org/terracotta/quartz/DefaultClusteredJobStore.java
```diff
@@ -825,6 +825,33 @@ public void clearAllSchedulingData() throws JobPersistenceException {
...
+    lock();
+    try {
+      TriggerWrapper tw = triggerFacade.get(key);
+
+      // does the trigger exist?
+      if (tw == null) { return; }
+
+      tw.setState(TriggerState.WAITING, terracottaClientId, triggerFacade);
+
+      applyMisfire(tw);
```
zemian: Again, why we need to call applyMisfire?

marwtki:
@jhouserizer I have submitted a signed agreement. Thank you!

jhouserizer:
CLA received.  Thank you for contributing!

quartz-core/src/main/java/org/quartz/core/QuartzScheduler.java
jhouserizer: I agree - I wonder if the method name should also be more explicit about what it does / is for.   Like what if you call this method on a trigger that is in BLOCKED state (and for proper reason is in blocked state)?

quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
jhouserizer: agreed.

quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
jhouserizer: agreed.

jhouserizer: I also wonder if this should only work for triggers in ERROR state.   (vs.  acquired, blocked, paused).

terracotta/bootstrap/src/main/java/org/terracotta/quartz/DefaultClusteredJobStore.java
jhouserizer: It is superfluous - will get detected and handled later if the misfire condition exists, but it is also likely that the trigger has actually misfired, so there is a small argument for eagerly updating it here (and the other locations).

quartz-core/src/main/java/org/quartz/simpl/RAMJobStore.java
jhouserizer: on point 1.  I agree the call is redundant/superfluous.  no real harm, no real gain.

on point 2. I agree that consistency would be good - it may make a difference at some point in the future.

jhouserizer:
Another finding:  simply resetting to "WAITING" state is a bit naive.    If the trigger group is paused, it should go to PAUSED, for example.

jhouserizer:
@marwtki ,  would you have a chance in the next few days to make some updates to this PR (based upon comments here) - or should I take your changes, edit, and create a new PR?

(I say "few days" because we want to release 2.3.0 in a week).

jjudd:
@jhouserizer, I'm not sure @marwtki is available to help with this anymore. She wrote it when working at Lucid last summer, but is no longer working with us. A couple months ago I applied @zemian's feedback and made a PR to her branch, but it was never merged. What would be easiest is for me to fork, apply the changes and feedback, and then I can open a new PR. 

I can do that either today or tomorrow if that works for you. If you would prefer to take the changes and make a pr yourself, feel free. Whatever is easiest for you.

jhouserizer:
Thanks @jjudd !  it would be appreciated if you did that.   

The unfortunate thing is that we'll then also need a CLA from you (see comment above from Oct 11, 2016 for instructions) - hopefully would only take you a few minutes to do.   (Or remind me if you have already done one in the past).

jjudd:
@jhouserizer, I have a branch here with the feedback applied: jjudd/quartz@8fec020adca47c232de52b6948536637cdb8d35e

I'm finding out if anyone has signed the CLA and can contribute the changes. Otherwise, I'm not sure I will be able to have the CLA reviewed and submitted before your release. If that is the case, feel free to use that branch as inspiration for your edits when making a new PR.

jhouserizer:
Thanks for the updates.  I'll take these changes and modify them to account for remaining comments above.

jhouserizer:
See new PR #125 .

Thanks to you all for the contributions.

PR #28, opened 2016-06-17
Simplify if-else statements in org.quartz.CronExpression.

garydgregory:
Statement unnecessarily nested within else clause.

### DIFF
--- a/quartz-core/src/main/java/org/quartz/CronExpression.java
+++ b/quartz-core/src/main/java/org/quartz/CronExpression.java
@@ -805,14 +805,12 @@ public final class CronExpression implements Serializable, Cloneable {
                     addToSet(val, end, v3, type);
                     i = vs.pos;
                     return i;
-                } else {
-                    addToSet(val, end, v2, type);
-                    return i;
                 }
-            } else {
-                addToSet(val, end, 1, type);
+                addToSet(val, end, v2, type);
                 return i;
             }
+            addToSet(val, end, 1, type);
+            return i;
         }
 
         if (c == '/') {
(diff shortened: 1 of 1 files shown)


jhouserizer:

garydgregory:
I emailed my CLA to tc-oss@softwareag.com.

jhouserizer:
Signed CLA received. Thanks!

zemian:
Hi @garydgregory, thanks for the PR. I have reviewed the changes and I am not sure it's this change is "simplifying" with the else block. The entire method is written in the style of using 'return' to exit method early if condition is met, hence the else is not used. This change is also making other part of if statements inconsistent in the method. Can you give me a better reason for the change?

garydgregory:
Hi @zemian,

My general feeling here is that code is easier to read when it is not unnecessarily nested.

For example, there is a lot of boilerplate in:

```
if (someTest) {
  return expr1;
} else {
  return expr2;
}
```

But less in:

```
if (someTest) {
  return expr1;
} 
return expr2;
```

Java is verbose enough as it is and I do not see the need to add to it by making the logic more deeply nested than it has to be.

This is a matter of personal style I suppose. 

Gary

zemian:
Hi Gary, the CronExpression logic is pretty involve, and I would hesitate to change unless there is a good reason with good test cases.

garydgregory:
The logic has not changed, the code behaves the same way, there is less
syntactic boiler plate. IOW, what add " } else {"'s if you do not have to?
ConExpressionTest passes just the same. Again, might just be personal
style, but I do believe that the code reads better and should be easier to
maintain.

Gary

zemian:
I am rejecting and closing this PR. As mentioned earlier, this proposed change actually affect the if/else condition of the logic and it's not fixing any problem, and I do not see it is any improvement on readability.

Issue #60, opened 2016-10-12
Quartz test failed with java8


zemian:
Setup:
```
Zemians-Air:quartz-test zemian$ mvn -version
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T11:41:47-05:00)
Maven home: /Users/zemian/apps/apache-maven-3.3.9
Java version: 1.8.0_60, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_60.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.11.6", arch: "x86_64", family: "mac"
```

Output:
```
Results :

Failed tests:   testFired(org.terracotta.quartz.upgradability.serialization.CalendarIntervalTriggerImplSerializationTest): Unrecognized serialized form saved to /Users/zemian/javadev/zemian/quartz-test/quartz/CalendarIntervalTriggerImpl-2.ser
  testConstructed(org.terracotta.quartz.upgradability.serialization.CalendarIntervalTriggerImplSerializationTest): Unrecognized serialized form saved to /Users/zemian/javadev/zemian/quartz-test/quartz/CalendarIntervalTriggerImpl-3.ser
  testFired(org.terracotta.quartz.upgradability.serialization.CronTriggerImplSerializationTest): Unrecognized serialized form saved to /Users/zemian/javadev/zemian/quartz-test/quartz/CronTriggerImpl-2.ser
  testConstructed(org.terracotta.quartz.upgradability.serialization.CronTriggerImplSerializationTest): Unrecognized serialized form saved to /Users/zemian/javadev/zemian/quartz-test/quartz/CronTriggerImpl-3.ser
  testFired(org.terracotta.quartz.upgradability.serialization.DailyTimeIntervalTriggerImplSerializationTest): Unrecognized serialized form saved to /Users/zemian/javadev/zemian/quartz-test/quartz/DailyTimeIntervalTriggerImpl-2.ser
  testConstructed(org.terracotta.quartz.upgradability.serialization.DailyTimeIntervalTriggerImplSerializationTest): Unrecognized serialized form saved to /Users/zemian/javadev/zemian/quartz-test/quartz/DailyTimeIntervalTriggerImpl-3.ser
[...]

PR #62, opened 2016-10-14
Add Maven wrapper

vpavic:
This PR adds [Maven wrapper](https://github.com/takari/maven-wrapper) to simplify project setup.


zemian:
Hi Vedran, I see you set warpper to use maven 3.3.9. The problem with this is that current Quartz is still on JDK6 compatible, and only maven 3.2.5 or below is able to run with JDK6. We should default to that maven version til we can move up the JDK.

vpavic:
Hey @zemian, that completely slipped my mind - I've updated the PR.

zemian:
Vedran, Unfortunately I think the wrapper itself requires min of JDK7?

Also I think mvnw.cmd line 124 has an invalid '#' char. I needs to be replace by '@REM'

```
mvnw.cmd -version
'#' is not recognized as an internal or external command,
operable program or batch file.
Exception in thread "main" java.lang.UnsupportedClassVersionError: org/apache/maven/wrapper/MavenWrapperMain : Unsupported major.minor version 51.0
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClassCond(ClassLoader.java:631)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:615)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:141)
        at java.net.URLClassLoader.defineClass(URLClassLoader.java:283)
        at java.net.URLClassLoader.access$000(URLClassLoader.java:58)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:197)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:301)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
Could not find the main class: org.apache.maven.wrapper.MavenWrapperMain.  Program will exit.
```

zemian:
We prob should move away from JDK6 soon anyhow, since it's pretty old. But I image that will be a major release bump. If wrappers does not support JDK6, then wait til we create a new branch for JDK7 for later inclusion. As I do like the maven wrapper tool and it's benefit.

vpavic:
Hey @zemian, you're right, Takari Wrapper indeed does require at least JDK7.

However, after giving this some thought, Quartz's compatibility with Java 6 doesn't necessarily need to affect the JDK version used to build the project meaning we could move the project build itself to newer version (preferably JDK8) but retain the runtime compatibility with Java 6.

In order to do so we should:

- fix #60 (fails for me too)
- use Animal Sniffer to ensure compatibility with Java 6
- add Wrapper

Your thoughts on this?

Also the invalid '#' char is Takari bug that was fixed in takari/maven-wrapper#27, I'll check why the script was generated using the old version.

zemian:
I think mixing different versions of tool and main code is confusing. I like push Quartz code to be JDK7 compatible instead. I will drop a mail to dev-list to see if all agree to make 2.3.x start using JDK7 instead.

vpavic:
:+1: from me for bumping the minimum required version to Java 7.
Regarding the project build, Spring projects have been built using JDK 8 while maintaining compatibility down to Java 6 for some time now so it's a very common and tested scenario. You could at least consider it for Quartz too.

Since this PR is blocked until the project gets built by JDK 7+ anyway, I'll update the Maven Wrapper config to target 3.3.9 again and fix the invalid '#' char issue. Is that OK with you?

zemian:
Yep, thanks @vpavic !

Issue #24, opened 2016-06-10
Null/zero length trigger job_data blob - JobDataMap lost


j1mcross:
Quartz 2.2.1
OracleDelegate

We have a number of triggers for a single job with the execution parameters stored in the JobDataMap. Once the trigger is created this data appears to be stored in the job_data column of triggers.

We do not change this data. We occasionally add and remove triggers but the data is static once the trigger is created.

The job is not marked with @PersistJobDataAfterExecution annotation.

Sporadically, for no obvious reason, the data in this column vanishes and the trigger will not fire and attempts to read the trigger information will fail.

Upon investigation it appears that the OracleDelegate is rewriting the data in this column always every time the trigger is accessed for scheduling of misfire handling. It rewrites by updating with an empty blob then updating with the data. It seems the first part of this is working but the resaving of the data is for some reason fails occasionally with the resulting problem of a zero length blob and missing JobDataMap.

First, since the job is not marked to persist after execution, it should not be trying to persist the data after execution. It also should not be persisting the data in any other handling.

Second, this writing of the JobDataMap appears to be controlled by the JobDataMap dirty flag. When the trigger is created the JobDataMap is serialized with the dirty flag set to true. When the trigger is retrieved the JobDataMap is retrieved with the flag true and the flag is never cleared until after the update of the blob occurs.

Setting false the dirty flag after retrieval of the JobDataMap, assuming that would have no other adverse consequences, might address this issue.

zemian:
I have verified this is a bug with test case now.

Instead of resetting the dirty flag everywhere where we load it from DB, I think the by definition of the JobDataMap constructor that takes an existing map should not mark dirty flag as true. This small change should fix the problem.

zemian:
This is already fixed.

PR #73, opened 2016-10-25
Fix JobDataMap dirty flag when constructing from existing map

zemian:
When loading JobDataMap from DB, we use the constructor wiht existing map, however
we did not clear the dirty flag when this happens and cause Trigger updating job
data unnecessary, and even can cause truncated data in raced condition case.

Issue #24


quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/StdJDBCDelegate.java
```diff
@@ -1158,9 +1158,9 @@ public int updateTrigger(Connection conn, OperableTrigger trigger, String state,
             JobDetail jobDetail) throws SQLException, IOException {
 
         // save some clock cycles by unnecessarily writing job data blob ...
-        boolean updateJobData = trigger.getJobDataMap().isDirty();
+        boolean updateJobData = trigger.getJobDataMap().isDirty() && (trigger.getJobDataMap().size() > 0);
```
jhouserizer: Regarding the size > 0 check -- what if the size used to be > 0, and now the size is 0 because entries were removed, should not the change (the emptiness) be persisted?

zemian: Yeah, good catch! I think just the isDirty() check is enough without the size. I will update the PR when I get a chance. Thanks!

PR #80, opened 2016-11-07
Avoid of a classloader memory leak on redeploy of web applications

rkashaev:
I've figured out when using quartz scheduler on web applications and using redeploy of your application, a classloader memory leak appears. The cause in details:
1. Create schedulers, the plugin registers `new Thread` in runtime shutdown hooks.
2. When you shut down your server, it's OK, JVM finishes, and there is no any problem.
3. BUT if you redeploy your application, ... bla-bla-bla, you create schedulers again, and they register **another** `new Thread` in runtime shutdown hooks. JVM runtime shutdown hook collection is a GC-root. It holds your new Thread and it's classloader (WebAppClassLoader on Tomcat) with all consequences.
4. Every time you redeploy your application, you don't unload your WebAppClassLoader with thousands of loaded classes and this can cause a permgen memory leak.
The issue is actual in all versions of the plugin.
I've fixed it with a pretty simple code and added a test to check it. Also added Mockito dependency (you can drop it if you don't need).

### DIFF
--- a/quartz-plugins/pom.xml
+++ b/quartz-plugins/pom.xml
@@ -58,7 +58,13 @@
       <artifactId>junit-dep</artifactId>
       <version>4.8.2</version>
       <scope>test</scope>
-    </dependency>    
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.2.12</version>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>asm</groupId>
       <artifactId>asm</artifactId>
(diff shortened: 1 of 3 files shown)


zemian:
Hello @rkashaev thanks for the PR feedback. Though there is no harm in your change, however the changes is not very significant. The reason is that `ShutdownHookPlugin` is intended to be use in a SE env. For Web env like what you described, you should be shutting down quartz through the servlet's `ContextListner` destroy method lifecycle since as you describe, redeployment is not a "shutdown" use case. See also `org.quartz.ee.servlet.QuartzInitializerListener` for more details.

rkashaev:
So, I suggest just not to use ShutdownHookPlugin in a web environment?

zemian:
That's correct. Use QuartzInitializerListener for web application. If you have multiple schedulers instances in a web application, then either they all have to be in a same cluster, or each must use a unique scheduler name and a distinct store.

Issue #32, opened 2016-07-03
RAMJobStore fails to resume when paused without triggers


xtianus:
In Quartz 2.2.3, if I call 

`scheduler.pauseTriggers(GroupMatcher.triggerGroupEquals("abc"));`

on a scheduler without triggers, then follow it with a

`scheduler.resumeTriggers(GroupMatcher.triggerGroupEquals("abc"));`

the paused group "abc" is not removed from the list of paused groups in the scheduler, so that any subsequent Trigger that is added will be in paused state.
I don't know if this is the expected behavior, but it goes against logic.

You can see the problem in RAMJobStore.resumeTriggers():

```
Set<String> groups = new HashSet<String>();
synchronized (lock) {
Set<TriggerKey> keys = getTriggerKeys(matcher);
for (TriggerKey triggerKey: keys) {
                groups.add(triggerKey.getGroup());
                if(triggersByKey.get(triggerKey) != null) {
                    String jobGroup = triggersByKey.get(triggerKey).jobKey.getGroup();
                    if(pausedJobGroups.contains(jobGroup)) {
                        continue;
                    }
                }
                resumeTrigger(triggerKey);
}
for (String group : groups) {
                pausedTriggerGroups.remove(group);
}
}
```

As you can see, pausedTriggerGroups.remove(group) is called only if  getTriggerKeys(matcher) returns some elements, which doesn't happen when there are no triggers.

xtianus:
I guess it also happens in the more general use case of pausing a group (with or without triggers defined), removing all triggers, resuming the group (which will not be resumed) then adding a trigger (which will stay paused).

zemian:
Hi xtianus, thanks for providing the feedback. 
We need to verify this. Can you help provide a junit test case?

zemian:
This is now fixed.

PR #84, opened 2016-11-12
Fix RAMJobStore fails to resume when paused without triggers

zemian:
Fix Issue #32


zemian:
@jhouserizer can you please do a code review on this PR?

PR #88, opened 2016-11-16
Quartz 2 2 x

srampv:
missing trigger from simple_triggers table


srampv:
retrieveTrigger method in JobStoreSupport supposed to return null . but it throws JobPersistenceException if trigger is already present in triggers table . this could block other triggers to be recovered . loop breaks due to exception

zemian:
This Pull Request is invalid. Please stop creating useless Pull Request. If you have question using Quartz, please try the mailing list.

PR #115, opened 2017-03-28
interrupt monitor plugin added

ramaraochavali:


quartz-plugins/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        return this.DEFAULT_MAX_RUNTIME;
+    }
+
+    public void setDefaultMaxRunTime(long defaultMaxRunTime) {
+        this.DEFAULT_MAX_RUNTIME = defaultMaxRunTime;
+    }
+
+    // Trigger Listener Methods
+    public String getName() {
+        return "JobInterruptMonitorListner";
```
jhouserizer: Name shouldn't be hard-coded, should be obtained/remembered from the value passed in on the initialize method.

(also there's a typo in the spelling of this hard-coded name)

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/PluginConstants.java
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.
+ */
+package org.quartz.plugins.interrupt;
+
+/**
+ * The Class PluginConstants.
+ * 
+ * @author Rama Chavali
+ */
+public class PluginConstants {
```
jhouserizer: Is there a good reason to have these constants defined on a different class than the plugin class itself?

My preference would be for them to be on the plugin class.  (that matches other plugins).   Or at least the name of this class should be more complete (i.e. prefixed with JobInterruptMonitor) - and perhaps it should even be an interface.

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+ * enabled.
+ * 
+ * @see org.quartz.Scheduler#interrupt()
+ * 
+ * @author Rama Chavali
+ */
+
+public class JobInterruptMonitorPlugin extends TriggerListenerSupport implements SchedulerPlugin {
+
+    private static final String JOB_INTERRUPT_MONITER_KEY = "JOB_INTERRUPT_MONITER_KEY";
```
jhouserizer: "MONITOR"   (spelling typo)

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        protected Logger getLog() {
+            return log;
+        }
+
+        @Override
+        public void run() {
+            try {
+
+                // Interrupt the job here - using Scheduler API that gets propagated to Job's interrupt
+                getLog().info("Interrupting Job " + jobKey.getName() + ":" + jobKey.getGroup());
```
jhouserizer: I wonder if it would be nice to add some words about why it is being interrupted (configured max run time exceeded).

jhouserizer:
Hi @ramaraochavali  -

Overall this is looking pretty good.  Please see my few comments/suggestions.

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
ramaraochavali: ok. will make this change

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/PluginConstants.java
ramaraochavali: ok. moved them to the plugin class it self

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
ramaraochavali: good catch..changed

quartz-plugins/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
ramaraochavali: changed the text

jhouserizer:
Thanks for the updates . Looks good.

Issue #71, opened 2016-10-21
Quartz causes 120 to 1500 connection attempts per second when DB is not available


oberlies:
When an JobPersistenceException is caught in the main QuartzSchedulerThread loop, the loop continues and immediately retries to read the triggers from the database. This leads to 120 to 1500 database connection attempts per second (!!). This leads to a flood of error log entries on our server.

Quartz should automatically throttle the retry attempts in case of database connection errors.

The first exception (and only one logged by Quartz directly) is

```
2015 10 10 18:58:06#+00#ERROR#org.quartz.core.ErrorLogger##anonymous#global_QuartzSchedulerThread#na#services#dispatcher#web##An error occurred while scanning for the next triggers to fire.
org.quartz.JobPersistenceException: Failed to obtain DB connection from data source 'DefaultDB': java.sql.SQLException: Could not retrieve datasource via JNDI url 'java:comp/env/jdbc/DefaultDB' <exception details from driver>. [See nested exception: java.sql.SQLException: Could not retrieve datasource via JNDI url 'java:comp/env/jdbc/DefaultDB' <exception details from driver>.]
    at org.quartz.impl.jdbcjobstore.JobStoreSupport.getConnection(JobStoreSupport.java:778)
    at org.quartz.impl.jdbcjobstore.JobStoreTX.getNonManagedTXConnection(JobStoreTX.java:71)
    at org.quartz.impl.jdbcjobstore.JobStoreSupport.executeInNonManagedTXLock(JobStoreSupport.java:3784)
    at org.quartz.impl.jdbcjobstore.JobStoreSupport.acquireNextTriggers(JobStoreSupport.java:2756)
    at org.quartz.core.QuartzSchedulerThread.run(QuartzSchedulerThread.java:272)
Caused by: java.sql.SQLException: Could not retrieve datasource via JNDI url 'java:comp/env/jdbc/DefaultDB' <exception details from driver>.
    at org.quartz.utils.JNDIConnectionProvider.getConnection(JNDIConnectionProvider.java:163)
    at org.quartz.utils.DBConnectionManager.getConnection(DBConnectionManager.java:108)
    at org.quartz.impl.jdbcjobstore.JobStoreSupport.getConnection(JobStoreSupport.java:775)
    ... 4 more|
```

---

This issue was originally reported and discussed here: https://jira.terracotta.org/jira/browse/QTZ-491

zemian:
[...]

PR #116, opened 2017-03-28
Delay reading from job store on repeated errors

whilemitja:
Hello,
this change helps with issue https://github.com/quartz-scheduler/quartz/issues/71, which is occasionally causing us some trouble in production.

If it's acceptable, it'd be great if we could also get it backported to the 2.2.x series, so that we don't have to wait until 2.3 is finished.. I don't know how that is done, though..

If it's not acceptable, please let me know of a better approach and I'm happy to change the implementation..

### DIFF
--- a/quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
+++ b/quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
@@ -27,8 +27,6 @@ import org.quartz.JobPersistenceException;
 import org.quartz.SchedulerException;
 import org.quartz.Trigger;
 import org.quartz.Trigger.CompletedExecutionInstruction;
-import org.quartz.impl.jdbcjobstore.JobStoreSupport;
-import org.quartz.spi.JobStore;
 import org.quartz.spi.OperableTrigger;
 import org.quartz.spi.TriggerFiredBundle;
 import org.quartz.spi.TriggerFiredResult;
@@ -269,7 +267,7 @@ public class QuartzSchedulerThread extends Thread {
                 // wait a bit, if reading from job store is consistently
                 // failing (e.g. DB is down or restarting)..
                 if (acquiresFailed > 1) {
-                    long delay = computeDelayForRepeatedErrors(qsRsrcs.getJobStore(), acquiresFailed);
+                    long delay = computeDelayForRepeatedErrors(acquiresFailed);
                     try {
                         Thread.sleep(delay);
                     } catch (InterruptedException ignore) {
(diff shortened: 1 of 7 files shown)


jhouserizer:
Hi @whilemitja , 

Thanks!

whilemitja:
CLA submitted..

pierrel:
@jhouserizer any update on this?

quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
```diff
@@ -27,6 +27,8 @@
 import org.quartz.SchedulerException;
 import org.quartz.Trigger;
 import org.quartz.Trigger.CompletedExecutionInstruction;
+import org.quartz.impl.jdbcjobstore.JobStoreSupport;
```
jhouserizer: The direct dependencies on a given JobStore implementation is running afoul of intended abstraction and isolation of the code organization / design.

I think we need to consider a new abstraction for this fix that doesn't create a direct dependency on JobStoreSupport.

quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
```diff
@@ -424,6 +440,23 @@ public void run() {
...
+    private static long computeDelayForRepeatedErrors(JobStore jobStore, long acquiresFailed) {
+        long interval;
+
+        if (jobStore instanceof JobStoreSupport) {
+            interval = ((JobStoreSupport)jobStore).getDbRetryInterval();
+        } else {
+            interval = 250 * acquiresFailed;
+        }
+
+        if (interval > 5000)
```
jhouserizer: If the dbRetryInterval is 15 seconds, this will reduce the delay to 5 seconds.  I wonder if we'd prefer to keep the 15 seconds?   I understand that in the "else" case above, where 250 is multiplied by aquiresFailed that the interval could get huge, and therefore need to be constrained to a maximum as is being done here, but maybe that should only apply in the "else" case above.

quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
jhouserizer: What about adding a method to the JobStore interface (which we can do with 2.3.0 - but not 2.2.x) named something like "getAquireRetryDelay()"  and the implementation on JobStoreSupport could return the value of dbRetryInterval, and the implementation on RAMJobStore could return 50 or something small, and  etc.

whilemitja: I see it more like an integration-with, rather than dependency-on, but ok, will remove.

whilemitja:
Ah, now I remember what the initial trigger for pulling in JobStoreSupport was - (at least) one of the tests sets the retry interval to something very small (5 or 50ms), then fails if the test takes too long to finish.. With the new/extra delays, the test indeed took too long, and so I thought it'd be better to take the retry interval into account rather than relaxing the test's time limit..

I see that 2.3.0 release is planned for April 16th, which is fairly soon, so I'll go ahead and add that method you suggested to JobStore..

jhouserizer:
This looks nice. Thanks again for your contribution!

petraBlock:
could this change be backported to the 2.2.x series as this causes us a lot of trouble in production? 
or is the 2.3.0 an official stable release? I  was a bit puzzled as it is not mentioned at all on the http://www.quartz-scheduler.org site.. furthermore I could not find any release note explaining the changes between 2.2.x and 2.3.0.
thanks.

dop89:
Any update on this? When will version 2.3 be released?

PraveenSrDeveoloper:
Hi all,
I was also facing the same issue ,as a Temp solution I have placed this
<validation>
  <check-valid-connection-sql>select 1 from dual</check-valid-connection-sql>
  <validate-on-match>true</validate-on-match>
  <background-validation>false</background-validation>
</validation>

in  <datasource> .

sugyanipatnaik:
Praveen , 

Can you tell me where exactly you did the changes, which file?

PR #152, opened 2017-06-14
- RecurrenceTrigger with iCalendar rfc5545 recurrence rules added.

elfogre:
I have a need to use recurrence rules to define quartz jobs. I read https://jira.terracotta.org/jira/browse/QTZ-252 discussion about it and decide to implement over the solution suggested in the issue. I've also included tests for the new trigger.


elfogre:
Forgot to push commit changes... close

Issue #176, opened 2017-08-01
@DisallowConcurrentExecution jobs are running at same time on different nodes when cluster de-synced


pavelkokush:
**Use case:** Quartz cluster with node A and B. All jobs are marked as @DisallowConcurrentExecution. Node A detect that node B "failed", because node B did not checkin for some period of time. But node B is ok (just failed to check-in for some reason) and there is one running job on node B. 

**Actual:** node A will change status from STATE_BLOCKED to STATE_WAITING for all jobs that was started on node B. So jobs will start running on node A even if same jobs are still running on node B. It is **NOT EXPECTED**, because jobs are marked as @DisallowConcurrentExecution!!!

**Expected:** Probably add some config like "org.quartz.jobStore.pauseOnNodeFail: true". Then pause jobs those have status STATE_BLOCKED. So user that is monitoring cluster will verify cluster state manually and decide is ok to un-pause job, then un-pause it manually.

pavelkokush:
It is in `JobStoreSupport.clusterRecover()` 
```
// free up stateful job's triggers
if (ftRec.isJobDisallowsConcurrentExecution()) {
    getDelegate()
            .updateTriggerStatesForJobFromOtherState(
                    conn, jKey,
                    STATE_WAITING, STATE_BLOCKED);
    getDelegate()
            .updateTriggerStatesForJobFromOtherState(
                    conn, jKey,
                    STATE_PAUSED, STATE_PAUSED_BLOCKED);
}

```

zemian:
Hi, sorry for the long delay in respond to this issue. Your issue seems to be duplication of #107 ! Please try the solution provided over there. Check carefully about your cluster setting in scheduler config and your DB isolation level. If your problem is exists, try the `acquireTriggersWithinLock=true` config as discussed over there.

PR #178, opened 2017-08-07
pause running jobs on failed node if pauseTriggeredJobOnFailedNode flag set #176

pavelkokush:


quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
```diff
@@ -3575,6 +3587,17 @@ protected void clusterRecover(Connection conn, List<SchedulerStateRecord> failed
...
                                             conn, jKey,
                                             STATE_PAUSED, STATE_PAUSED_BLOCKED);
+
+                            if (pauseTriggeredJobOnNodeFail) {
+                                if (ftRec.getJobKey() != null) {
+                                    log.warn(String.format("job %s paused, because it was running when node %s failed. Make sure that job is NOT running, then unpause it", ftRec.getJobKey(), ftRec.getFireInstanceId()));
+
+                                    List<OperableTrigger> triggers = getTriggersForJob(conn, ftRec.getJobKey());
+                                    for (OperableTrigger trigger : triggers) {
+                                        pauseTrigger(conn, trigger.getKey());
```
jontejj: Should it catch exceptions and continue to try pausing the next trigger?

pavelkokush: then it will be inconsistent state: some triggers paused, some not. Also then we will have same job running at same time on diff nodes. So better to rollback full transaction, as it now?

jontejj:
Aha! So the transaction is not inside this method. Good!

zemian:
Hi, as I stated in your original issue #176, that this is currently working as expected that you might need additional config change. I will close this PR as invalid for now.

Issue #192, opened 2017-10-19
JobExecutionContext can't obtain correct recoveringTriggerKey


liyuankui:
some bug in the internal logic

zemian:
This issue is now fixed

PR #193, opened 2017-10-19
refs #192 job execution context impl returns correct recovering job key

liyuankui:
TriggerKey is always (name,group) pair, name first.


zemian:
Hi @liyuankui Thank you for the PR! I have created a new PR#349 with your contribution against 2.3.x branch. Will close this one as dup.

Issue #206, opened 2018-01-19
How to customize JobRunShellFactory


gabrywu:
When I study StdSchedulerFactory source code and found JobRunShellFactory can only JTAJobRunShellFactory or JTAAnnotationAwareJobRunShellFactory,  JobRunShellFactory specific  can not be configured in the configuration file. why ？

PR #207, opened 2018-01-19
refs #206 How to customize JobRunShellFactory

gabrywu:
fix refs #206 
able to customize JobRunShellFactory

### DIFF
--- a/quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java
+++ b/quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java
@@ -285,6 +285,7 @@ public class StdSchedulerFactory implements SchedulerFactory {
 
     public static final String MANAGEMENT_REST_SERVICE_HOST_PORT = "org.quartz.managementRESTService.bind";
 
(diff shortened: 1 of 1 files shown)


quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java
```diff
@@ -1220,14 +1221,30 @@ else if (schedInstId.equals(SYSTEM_PROPERTY_AS_INSTANCE_ID)) {
     
             if (userTXLocation != null) {
                 UserTransactionHelper.setUserTxLocation(userTXLocation);
+                if (wrapJobInTx) {
```
jarlah: You havent just added some more code here. You have changed the previous behaviour of this class by moving the if(wrapJobInTx) etc inside if(userTXLocation).... is this is a bug fix? If not, how you can be sure that this change does not affect old behaviour?

zemian:
Hi, Thank you for the PR contribution! 

As @jarlah have already commented, please fix the proper code for contribution. We would appreciated if you can add full unit tests as well.

Because this PR is a feature request, you would need to sign the Terracotta Contribution License Agreement. Please find the form here: http://www.quartz-scheduler.org/community/contribute.html

PR #212, opened 2018-02-21
QuartzInitializerListener: fix a typo

caladyon:
It seems that the shutdownWaitPref variable should be null-checked instead of shutdownPref variable, that was checked few rows before.


zemian:
@caladyon Thank you for the fix! I have ported your change into a new PR ##322 against quartz-2.3.x branch instead.

PR #214, opened 2018-03-01
JobBuilder.storeDurably() to reuse JobBuilder.storeDurably(boolean)

codingsince1985:
DRY and let JobBuilder.storeDurably() reuse JobBuilder.storeDurably(boolean)


zemian:
Thank you for the PR! I agree that this can be a improvement.

This change should be merge against 2.3.x branch though. I have re-created a new PR #336 with your commit and included changelog update.

PR #245, opened 2018-07-12
Changed varchar length TRIGGER_NAME from 80 to 200

matzeeable:
After deploying quartz to a Sybase database the following error occured:

```
Error exec quartz initialisation com.sybase.jdbc4.jdbc.SybBatchUpdateException: JZ0BE: BatchUpdateException: Error occurred while executing batch statement: Column lengths of string type referencing and referenced columns don't match. referencing column = 'TRIGGER_NAME', referenced column = 'TRIGGER_NAME'. 
```

After researching I found out that [tables_sybase.sql](https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/tables_sybase.sql) is using `TRIGGER_NAME varchar(80)` instead of `TRIGGER_NAME varchar(200)`. Length should be equivalent to [tables_mysql.sql](https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/tables_mysql.sql).

Thanks to @iDschepe figuring this out.

### DIFF
--- a/quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/tables_sybase.sql
+++ b/quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/tables_sybase.sql
@@ -114,7 +114,7 @@ go
 
 create table QRTZ_CRON_TRIGGERS (
 SCHED_NAME varchar(120) not null,
-TRIGGER_NAME varchar(80) not null,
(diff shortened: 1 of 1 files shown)


jarlah:
this will not be backwards compatible?

PR #252, opened 2018-07-30
Fix XMLSchedulingDataProcessor to compile with Java 10

ebourg:
Hi,

The return type of the `getPrefixes​()` method  of `javax.xml.namespace.NamespaceContext` changed in Java 10 from `List` to `List<String>`. This causes a compilation error in XMLSchedulingDataProcessor. Here is a fix addressing this issue.

### DIFF
--- a/quartz-core/src/main/java/org/quartz/xml/XMLSchedulingDataProcessor.java
+++ b/quartz-core/src/main/java/org/quartz/xml/XMLSchedulingDataProcessor.java
@@ -195,7 +195,7 @@ public class XMLSchedulingDataProcessor implements ErrorHandler {
               return XMLConstants.NULL_NS_URI;
           }
         
-          public Iterator<?> getPrefixes(String namespaceURI)
+          public Iterator<String> getPrefixes(String namespaceURI)
           {
               // This method isn't necessary for XPath processing.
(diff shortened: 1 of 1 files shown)


jarlah:
added in into my fork

zemian:
Hi, thanks for submitting PR. Current master branch is using JDK8. We plan to support Java11 on the quartz-3.0.x in future, which already has this fixed.

Closing PR as invalid.

dreis2211:
Hi @zemian . Any (rough) ETA on when quartz-3.0 arrives? The changes made in 647d3f20bbe5e8970c6d6d2cdf3bd07345aa4ee2 seem to be also not too invasive. Maybe a backport to 2.x would be possible? At least it would be highly appreciated. After all JDK 12 is already released.

zemian:
Hi @dreis2211, unfortunately no plan on any releases. But probably won't be for a while since there is not much activities going on. If you need this in 2.x branches, you can file a feature request and let it get worked on by whoever have the time.

zemian:
I thought I answered it above: **no plan**.

zemian:
BTW, quartz-2.x does not use JDK 10, so there is no "backward" port of any sort. So I am not sure what's been requested also.

Issue #107, opened 2017-02-24
In clustered mode，if do not use acquireTriggersWithinLock properties will appear ABA problem. 


wenniuwuren:
In clustered mode，if do not use acquireTriggersWithinLock properties will appear ABA problem. 

here is the quartz source code:
```
public List<OperableTrigger> acquireNextTriggers(final long noLaterThan, final int maxCount, final long timeWindow)
        throws JobPersistenceException {
if(isAcquireTriggersWithinLock() || maxCount > 1) { 
            lockName = LOCK_TRIGGER_ACCESS;
        } else {
            lockName = null;
        }
...
```
```
protected List<OperableTrigger> acquireNextTrigger(Connection conn, long noLaterThan, int maxCount, long timeWindow)
        throws JobPersistenceException {
 
        ... 

        do {
            currentLoopCount ++;
            try {
                List<TriggerKey> keys = getDelegate().selectTriggerToAcquire(conn, noLaterThan + timeWindow, getMisfireTime(), maxCount);
                
                // No trigger is ready to fire yet.
                if (keys == null || keys.size() == 0)
                    return acquiredTriggers;

                long batchEnd = noLaterThan;

                for(TriggerKey triggerKey: keys) {
                    
                    ...

                    // We now have a acquired trigger, let's add to return list.
                    // If our trigger was no longer in the expected state, try a new one.
                    int rowsUpdated = getDelegate().updateTriggerStateFromOtherState(conn, triggerKey, STATE_ACQUIRED, STATE_WAITING);
                    if (rowsUpdated <= 0) {
                        continue; // next trigger
                    }
                    nextTrigger.setFireInstanceId(getFiredTriggerRecordId());
                    getDelegate().insertFiredTrigger(conn, nextTrigger, STATE_ACQUIRED, null);

                    if(acquiredTriggers.isEmpty()) {
                        batchEnd = Math.max(nextTrigger.getNextFireTime().getTime(), System.currentTimeMillis()) + timeWindow;
                    }
                    acquiredTriggers.add(nextTrigger);
                }

                ...

            } catch (Exception e) {
                throw new JobPersistenceException(
                          "Couldn't acquire next trigger: " + e.getMessage(), e);
            }
        } while (true);
        
        // Return the acquired trigger list
        return acquiredTriggers;
    }
```

[...]

PR #280, opened 2018-11-19
Fix quartz Job fired twice in cluster mode

wangfan9002:
`org.quartz.jobStore.acquireTriggersWithinLock=true`
if not configured, Quartz Job will appear fired twice in cluster mode, because JobStoreSupport file line 1284 ABA problem  [https://github.com/quartz-scheduler/quartz/issues/107](https://github.com/quartz-scheduler/quartz/issues/107)
Double check to resolve the ABA problem

### DIFF
--- a/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
+++ b/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
@@ -2970,6 +2970,20 @@ public abstract class JobStoreSupport implements JobStore, Constants {
                     trigger.getKey());
             if (!state.equals(STATE_ACQUIRED)) {
                 return null;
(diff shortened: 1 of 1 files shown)


zemian:
Hi, thank you for the PR contribution. However I do not think the changes is the best solution for this particular problem. Please see #107 comments for solutions.

quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
```diff
@@ -2970,6 +2970,20 @@ protected TriggerFiredBundle triggerFired(Connection conn,
                     trigger.getKey());
             if (!state.equals(STATE_ACQUIRED)) {
                 return null;
+            } else if(! isAcquireTriggersWithinLock()) {
```
monalisha1994, line 2973: This feature is not added in mongoDBJobStore

Issue #281, opened 2018-11-20
can not use properties validateOnCheckout


fish7790714:
when i use C3p0PoolingConnectionProvider it default use ComboPooledDataSource 。
at StdSchedulerFactory line 1392
` 
private void populateProviderWithExtraProps(PoolingConnectionProvider cp, Properties props) throws Exception {
        Properties copyProps = new Properties();
        copyProps.putAll(props);
        // Remove all the default properties first (they don't always match to setter name, and they are already
        // been set!)
        copyProps.remove(PoolingConnectionProvider.DB_DRIVER);
        copyProps.remove(PoolingConnectionProvider.DB_URL);
        copyProps.remove(PoolingConnectionProvider.DB_USER);
        copyProps.remove(PoolingConnectionProvider.DB_PASSWORD);
        copyProps.remove(PoolingConnectionProvider.DB_MAX_CONNECTIONS);
        copyProps.remove(PoolingConnectionProvider.DB_VALIDATION_QUERY);
        props.remove(PoolingConnectionProvider.POOLING_PROVIDER);
        setBeanProps(cp.getDataSource(), copyProps);
    }
`

"copyProps" not remove properties "validateOnCheckout",but ComboPooledDataSource  don't have this properties ,lead to  Exception
` 
Caused by: java.lang.NoSuchMethodException: No setter for property 'validateOnCheckout'
	at org.quartz.impl.StdSchedulerFactory.setBeanProps(StdSchedulerFactory.java:1447) ~[quartz-2.3.0.jar:na]
	at org.quartz.impl.StdSchedulerFactory.populateProviderWithExtraProps(StdSchedulerFactory.java:1405) ~[quartz-2.3.0.jar:na]
	at org.quartz.impl.StdSchedulerFactory.instantiate(StdSchedulerFactory.java:1040) ~[quartz-2.3.0.jar:na]
	... 74 common frames omitted
`

zemian:
What's your quartz.properties like on configuring these datasource?

zemian:
Close as not an issue

fish7790714:
I'm sorry that I didn't see your reply recently.
My  quartz.properties  is as follows:
```  java 
Properties prop = new Properties();
        // scheduler实例名称 随便取
        prop.put("quartz.scheduler.instanceName", "DefaultQuartzScheduler");
[...]

PR #305, opened 2018-12-28
Fix can not use properties 'validateOnCheckout'

fish7790714:
Hi,
 if use properties 'validateOnCheckout',it will throw exception 
``` java
Caused by: java.lang.NoSuchMethodException: No setter for property 'validateOnCheckout' at org.quartz.impl.StdSchedulerFactory.setBeanProps(StdSchedulerFactory.java:1447)
```
I've already described it in iusse #281 .
Now,I've fixed it.


zemian:
Thanks for the PR!

PR #340, opened 2019-02-13
Use all-caps table names in the liquibase script

immortaly007:
On case-sensitive databases, the table names in the Liquibase script cause issues, since they're all lower case. This change fixes that.

It also removes the `tdss` prefix from the table names.

### DIFF
--- a/quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/liquibase.quartz.init.xml
+++ b/quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/liquibase.quartz.init.xml
@@ -4,349 +4,349 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
 
     <changeSet id="quartz-init" author="nasazh">
-        <createTable tableName="qrtz_tdss_locks">
-            <column name="sched_name" type="VARCHAR(120)">
+        <createTable tableName="QRTZ_LOCKS">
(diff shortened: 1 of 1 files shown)


zemian:
Hi @immortaly007 Thank you for the PR! Would you mind providing some instruction as how to use this file? You can post them here as comment, and I will try to create a WIKI page so other can use it.

immortaly007:
Hi @zemian,

The file is useful for users that want to control their database schema using [Liquibase](https://www.liquibase.org/), as opposed to using the "quartz standard" way of generating/controlling the database schema. The advantage of using Liquibase is that this might integrate better with the existing schema generation, and it keeps the application and quartz schema generation in one place. 

The easiest way the file can be used is to copy it next to the users' existing changelog file (for example to `/src/main/resources/db/changelog/`), and then reference it in the users' existing changelog using the following line of XML:
```
<include file="liquibase.quartz.init.xml" relativeToChangelogFile="true" />
```

This way, Liquibase will ensure that the quartz tables are created (with their default names).

When using the `spring-boot-starter-quartz`, I'd recommend setting the following application property to prevent quartz from also generating/overriding the schema:
```
spring.quartz.jdbc.initialize-schema=never
```

zemian:
Thank you @immortaly007 for the info! Few updates I have for this PR:

1. Because we have to add these into 2.3.x, I have created a new PR #355 with your commits in it. So I will close this as dup.
2. I have learned few things about Liquibase and tested with Postgres. I found there is a need to modify the config. I also improved it with table_prefix.
3. I now have created a WIKI page describing how to use this tool: https://github.com/quartz-scheduler/quartz/wiki/How-To-Setup-Databases-for-Quartz#using-liquibase-tool

Issue #234, opened 2018-05-29
jobScheduled listener not triggered


claudiowilson:
When using this api:

`void scheduleJob(JobDetail jobDetail, Set<? extends Trigger> triggersForJob, boolean replace) throws SchedulerException;` the jobScheduled callback for the SchedulerListener does not get triggered. This is because when storing the triggers `QuartzScheduler` does not call `notifySchedulerListenersSchduled` in the API. I can put up a fix if you'd like, since it is relatively simple.

zemian:
Fixed with PR #351

PR #351, opened 2019-02-13
Miss notify SchedulerListeners in QuartzScheduler.java

zemian:


zemian:
This PR is for issue #234

PR #354, opened 2019-02-14
Use all-caps table names in the liquibase script #340

zemian:
Use all-caps table names in the liquibase script #340


zemian:
Bad merge in PR!

PR #401, opened 2019-02-27
Code Cleanups

gtiwari333:
This contains several code cleanup items based on analysis performed by Intellij Code Analysis.
 eg: 

- removal of unused imports
- removal of _ as variable name
- using lambdas where possible
- Using foreach loop instead of with index
- removal of type on right 
- using try-with-resources
- etc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -217,7 +217,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-checkstyle-plugin</artifactId>
-          <version>2.9.1</version>
+          <version>3.0.0</version>
           <configuration>
             <encoding>UTF-8</encoding>          
             <configLocation>${checkstyle-config-file}</configLocation>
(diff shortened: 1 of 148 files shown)


zemian:
Hi, thank you for the feedback. Your PR failed to build. Please see https://github.com/quartz-scheduler/quartz/wiki/ContributorGuide for guide.

Also, we are reluctant to simply reformat all codes for the sake of changes, Especially that modify logics and failed the build. We like to have a cleaner code at some point, but maybe in a smaller changeset at a time where it can be validated more safely.

gtiwari333:
It looks the failed build is due to the following  change of variable type from ArrayList to List on:

- org.quartz.impl.calendar.AnnualCalendar.excludeDays 

Failed reason: Unrecognized serialized form saved to /home/vsts/work/1/s/quartz/AnnualCalendar-0.ser

I can fix this build and submit again.

I agree that this contains a lot of changes. Let me know if you prefer the same updates broken out into smaller updates in multiple PRs.

zemian:
Close due to inactivities

PR #408, opened 2019-03-22
Updates to Lesson Links on Index.adoc

Alubaczewski:
Links to lessons were throwing 404's
Updated .html to .md to match file format
Links now properly route to the correct documents


zemian:
Hi @Alubaczewski thanks for the PR.

The links were using `.html` suffix on purpose since we publish the official docs onto http://www.quartz-scheduler.org/documentation. Some how the link will not auto rename using Jykell if we are to use `.md`.

PR #148, opened 2017-05-23
CronExpression add getTimeBefore function

todaychi:
We can use getTimeBefore to calculate the date which is last time with frequency rule; 
It's opposite from getTimeAfter;

PR #421, opened 2019-04-02
CronExpression.getTimeBefore() using getTimeAfter()

mykeul:
Here is an alternative to recently closed PR #148.
The changes can be applied to branch quartz-2.3.x too.

### DIFF
--- a/quartz-core/src/main/java/org/quartz/CronExpression.java
+++ b/quartz-core/src/main/java/org/quartz/CronExpression.java
@@ -29,6 +29,7 @@ import java.util.SortedSet;
 import java.util.StringTokenizer;
 import java.util.TimeZone;
 import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Provides a parser and evaluator for unix-like cron expressions. Cron 
@@ -199,6 +200,10 @@ import java.util.TreeSet;
 public final class CronExpression implements Serializable, Cloneable {
 
     private static final long serialVersionUID = 12423409423L;
+    private static final long[] GET_TIME_BEFORE_INTERVALS = new long[] {
+            TimeUnit.MINUTES.toMillis(1), TimeUnit.HOURS.toMillis(1), TimeUnit.DAYS.toMillis(1),
+            TimeUnit.DAYS.toMillis(7), TimeUnit.DAYS.toMillis(31), TimeUnit.DAYS.toMillis(366),
+            TimeUnit.DAYS.toMillis(3653), TimeUnit.DAYS.toMillis(36525) };
     
     protected static final int SECOND = 0;
     protected static final int MINUTE = 1;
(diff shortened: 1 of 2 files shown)


quartz-core/src/main/java/org/quartz/CronExpression.java
```diff
@@ -1586,11 +1591,22 @@ protected void setCalendarHour(Calendar cal, int hour) {
...
     /**
-     * NOT YET IMPLEMENTED: Returns the time before the given time
+     * Returns the time before the given time
      * that the <code>CronExpression</code> matches.
      */ 
     public Date getTimeBefore(Date endTime) { 
-        // FUTURE_TODO: implement QUARTZ-423
+        final Date fireTimeAfterOrigin = getTimeAfter(endTime);
+        if (fireTimeAfterOrigin == null) return null;
+        for (final long interval : GET_TIME_BEFORE_INTERVALS) {
```
chrisdennis: This looks like it wants to be two binary searches in opposite directions. The outer loop is essentially searching backwards in ever widening circles trying to establish a lower bound on the time we want. The inner loop is then searching forwards trying to find the time we want. It would seem that it would be more efficient and clearer if we made the first search use an actual exponential growth pattern, and then once we've established a lower and upper bound we can do a binary search from that point. The following does exactly that:
```java
    public Date getTimeBefore(Date endTime) {
        final Date after = getTimeAfter(endTime);

        if (after == null) return null;

        long afterMillis = after.getTime();

        long upper = afterMillis;
        long lower = Long.MIN_VALUE;

        //establish a lower bound (and refine the upper while we're at it)
        for (long i = 1L; i > 0; i <<= 1L) {
            long candidateMillis = afterMillis - i;
            Date candidate = getTimeAfter(new Date(candidateMillis));
            if (candidate == null) {
                return null;
            }
            if (candidate.equals(after)) {
                upper = candidateMillis - 1;
            } else {
                lower = candidateMillis;
                break;
            }
        }

        //do a binary search for the threshold value
        while (lower < (upper - 1)) {
            long middle = (lower + upper) >>> 1;

            Date candidate = getTimeAfter(new Date(middle));
            if (candidate == null) {
                return null;
            }

            if (candidate.equals(after)) {
                upper = middle - 1;
            } else {
                lower = middle;
            }
        }
        return getTimeAfter(new Date(lower));
    }
```

rsrini7:
Awaiting for this feature. When this will be available ? Thanks.

akomakom:
Hello! Thank you very much for your contribution and interest in helping improve the Quartz community.

After a period of dormancy, the Quartz project is back under steady maintenance by multiple volunteers, who are working to once again handle contributions such as yours.

We notice that your contribution was made without use of the DCO feature (the sign-off feature on commits via the -s option). Can you please update your PR with commits that use the -s option, agreeing to assign copyright ownership and other terms as described at the contributor agreement referenced here: https://github.com/quartz-scheduler/contributing/blob/main/CONTRIBUTING.md

You can easily add signoff to your previous commits by running (on your PR branch):
```shell
git commit --amend --signoff --no-edit 
git push -f
```

Or (for multiple commits):
```shell
# change 5 to the number of commits
git rebase HEAD~5 --signoff
git push -f
````

**Comment by mykeul** [comment]
Hello, thanks for volonteering. Signed-off my commit a few minutes ago: all automatic checks seem ok.
Thanks again.

File: quartz-core/src/main/java/org/quartz/CronExpression.java (hunk 1 above)
Comment (chrisdennis) [reply quartz-core/src/main/java/org/quartz/CronExpression.java:1600]: The above is overly precise and therefore too costly... please ignore.

**Review by chrisdennis — APPROVED** [review]
I think what we have here is good. I think I would prefer something like the following:
```java
public Date getTimeBefore(Date endTime) {
    final Date after = getTimeAfter(endTime);
    if (after == null) return null;

        long upper = after.getTime();
        long lower = 0;

        while (lower < (upper - 1)) {
            long middle = (((lower + upper) >>> 1L) + upper) >>> 1L;

            Date candidate = getTimeAfter(new Date(middle));
            if (candidate == null) {
                return null;
            }

            if (candidate.equals(after)) {
                upper = middle - 1;
            } else if (getTimeAfter(candidate).equals(after)) {
                return candidate;
            } else {
                lower = middle;
            }
        }
        return null;
   }
}
```
that said my preference isn't strong enough to change the code in this PR.

PR #443, opened 2019-05-09
Update tables_h2.sql to use IF NOT EXISTS for table creation

cemremengu:
Use `IF NOT EXISTS` to prevent initialization errors


chrisdennis:
Are we saying that H2 is special (none of the other db scripts use "if not exists" semantics) because it's often used in more of an adhoc manner or is this just the first stage in a proposal to move all the scripts to use "if not exists" semantics?

cemremengu:
I was working with H2 so that's why I only updated those. I can also send the updates for others if the team decides to move that way.

PR #456, opened 2019-06-12
liquibase improvement for oracle

slawekjaranowski:
some object name are too long for oracle db
valid names was copied from tables_oracle.sql


chrisdennis:
I'm struggling to see why Oracle is special here. If we merge this changeset aren't we inviting everyone to come in and point out how the schema in the liquibase file as it stands doesn't work for their specific database. It seems to me we're stuck with one of three situations:

1. We leave the liquibase file as is and let users customize for their DB via either the Liquibase API or a suitably crafted XSLT/XPath query.
2. We introduce a variant of the liquibase file for every DB for which we currently provide a sequence of DDL expressions.
3. We have a single liquibase file that constantly flits between all the forms above.

To me only 1 & 2 are sane states, and I'm leaning towards 1 being the most sane of our two choices.

slawekjaranowski:
If you stay with option 1 - leave the liquibase file as is - you should put somewhere information which DBMS is supported be this file.

Option 2 - we will have many files with a small count of differents. 

Option 3 - Now you have special "properties" for PostgreSql, so I thingk that one file with needed "properties" for diferent supported DBMS will be good.

slawekjaranowski:
If you still interest - please reopen.

Issue #474, opened 2019-08-07
StdSchedulerFactory ConcurrentModificationException reading system properties


davidmoten:
Using quartz 2.3.1 I get the following exception occasionally on startup of a webapp in Tomcat.

```
java.util.ConcurrentModificationException
        at java.util.Hashtable$Enumerator.next(Hashtable.java:1387)
        at java.util.Hashtable.putAll(Hashtable.java:523)
        at org.quartz.impl.StdSchedulerFactory.overrideWithSysProps(StdSchedulerFactory.java:495)
        at org.quartz.impl.StdSchedulerFactory.initialize(StdSchedulerFactory.java:472)
        at org.quartz.impl.StdSchedulerFactory.getScheduler(StdSchedulerFactory.java:1552)
```

This is due to `StdSchedulerFactory.overrideWithSysProps` which looks like this:

https://github.com/quartz-scheduler/quartz/blob/1ba8447fae062f241944bcc98ca30fa63a063304/quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java#L475-L496

By calling `putAll` the `entrySet()` iterator is  used on `System.getProperties()` which is not a thread-safe way of traversing System properties.

The fix would be to instead traverse the System properties by using the Enumeration `propertyNames()` and calling `getProperty` on each property name.

Happy to provide PR.

CC @SuperEventSteven

davidmoten:
I've submitted a PR #475 for this one

Issue #467, opened 2019-07-22
Security: XXE in initDocumentParser


krabbenpuler:
The method initDocumentParser() in the XMLSchedulingDataProcessor.java does not forbid DTDs, which allows a context-dependend attacker to perfom an [XXE](https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing). The vulnerability is confirmed working in version 2.2.3.

I reported this issue to the MITRE cooperation already which assigned the identifier [CVE-2019-13990](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-13990) to this vulnerability. Please confirm the existence of this vulnerability so that the CVE-entry can be completed and published.

The OWASP Foundation provides an [XXE Prevention CheatSheet](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html) that explains in detail the steps to prevent this security issue. 

Credits to Andreas Schoedl for working out the relevant code section affected by this issue.

jgallimore:
@krabbenpuler I've been taking a look at this, and I think I have a patch I can send as a PR for master. Do you have a specific test case I can verify it with?

krabbenpuler:
Sure, the following attack vector worked in my case: 

`<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE foo [<!ELEMENT foo ANY >
<!ENTITY xxe SYSTEM "/" >
]>
<job-scheduling-data xmlns="http://www.quartz-scheduler.org/xml/JobSchedulingData" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.quartz-scheduler.org/xml/JobSchedulingData http://www.quartz-scheduler.org/xml/job_scheduling_data_2_0.xsd" version="2.0">
<schedule>
<job>
<name>
xxe</name>
<group>
native</group>
<description>
&xxe;</description>
<job-class>
org.quartz.jobs.NativeJob</job-class>
<durability>
true</durability>
<recover>
false</recover>
</job>
</schedule>
</job-scheduling-data>
`

jgallimore:
That's great - thanks for that.

jgallimore:
PR for master: https://github.com/quartz-scheduler/quartz/pull/472

jgallimore:
I'll send an email, but I'm happy to file a CLA if needed.

jgallimore:
[...]

PR #475, opened 2019-08-07
Fix ConcurrentModificationException in StdSchedulerFactory

davidmoten:
Discussed in #474 

* Used `propertyNames()` to iterate System properties instead of the `entrySet()` (via `putAll`) to avoid a `ConcurrentModificationException`
* Added unit test for coverage of new code

I've not added a unit test that provoked a CME because it involves a race condition that is fiddly to provoke.

### DIFF
--- a/quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java
+++ b/quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java
@@ -65,6 +65,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.security.AccessControlException;
 import java.util.Collection;
+import java.util.Enumeration;
 import java.util.Locale;
 import java.util.Properties;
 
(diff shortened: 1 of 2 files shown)


davidmoten:
Are you around @zemian? I was hoping to get this simple PR reviewed and a release. Do you guys do quick turnaround on releases when bugs are discovered?

melloware:
@zemian doesn't seem to be responding as there is also a major security flaw that needs to be fixed.  https://github.com/quartz-scheduler/quartz/issues/467

quartz-core/src/main/java/org/quartz/impl/StdSchedulerFactory.java
```diff
@@ -492,7 +494,16 @@ private Properties overrideWithSysProps(Properties props) {
         }
 
         if (sysProps != null) {
-            props.putAll(sysProps);
+            // Use the propertyNames to iterate to avoid 
+            // a possible ConcurrentModificationException
+            Enumeration<?> en = sysProps.propertyNames();
```
chrisdennis: This is going to throw `ClassCastException` if any of the keys aren't strings, which is unfortunately possible here. We could instead use `stringPropertyNames()` which omits the non-String keys. Or, horror of horrors we could instead call `props.putAll((Properties) sysProps.clone())`

davidmoten: Quite right, thanks. I've fixed with the latest commit and didn't use `stringPropertyNames` because there is still a race condition where a value could be changed to an Object just after I've read `stringPropertyNames`.

davidmoten:
Ready for review again, thanks.

jhouserizer:
Hi @davidmoten ,

We appreciate your contribution!   It seems to fall into a rare gray area between changes that are trivial enough to not warrant a contributor's agreement, and ones that are significant enough to require one.   Would you mind filling out a contributor's agreement as outlined here: http://www.quartz-scheduler.org/community/contribute.html#submitting-contributions ?

davidmoten:
Sure, I'll fill out the contributor agreement.

davidmoten:
I've sent the agreement.

chrisdennis:
Sorry... I know this has taken longer than expected. @jhouserizer and I were both away on business last week... then I was ill, and now he's traveling. Once he gets back tomorrow we can check with the lawyers to make sure they've rubber-stamped the CA and then I can merge this.

chrisdennis:
@davidmoten I've cherry-picked your fix and the rebased as daggy and created a new PR under #512. I plan to merge assuming CI is clean. Let me know if anything looks amiss to you.

PR #481, opened 2019-08-16
Issue #467 provide XML parser with a strong configuration to prevent …

jgallimore:
…XXE attacks


quartz-core/src/main/java/org/quartz/xml/XMLSchedulingDataProcessor.java
```diff
@@ -173,7 +173,14 @@ protected void initDocumentParser() throws ParserConfigurationException  {
         docBuilderFactory.setAttribute("http://java.sun.com/xml/jaxp/properties/schemaLanguage", "http://www.w3.org/2001/XMLSchema");
         
         docBuilderFactory.setAttribute("http://java.sun.com/xml/jaxp/properties/schemaSource", resolveSchemaSource());
-        
+
+        docBuilderFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
```
davidmoten: I'm assuming that setting this feature is all that is required to make your unit test pass. Do you think it's worthwhile adding unit tests for the other features that are being set?

chrisdennis:
Replaced by the daggy PR #502

PR #490, opened 2019-09-12
Return at most maxCount triggers

bradbeck:
Addresses Issue: #491

In cases where the JDBC driver does not respect `PreparedStatement.setFetchSize` and `PreparedStatement.setMaxRows`, the current logic may return more than `maxCount` results.


chrisdennis:
Good catch. I've taken the liberty of rebasing your fix against the earliest commit (daggy fix: https://wiki.monotone.ca/DaggyFixes/) and added a unit test. That work is under a new PR #499

PR #493, opened 2019-09-17
Fix Javadoc mention a not existing method

jbarotin:


chrisdennis:
This looks like a fine change but I'd like to keep bug fixes daggy (https://wiki.monotone.ca/DaggyFixes/) so that they can be easily merged and tracked through the various branches. I rebased your commit against the earliest commit in GitHub (this is an old bug!) and created a new PR #498

PR #501, opened 2019-10-04
[master] Issue 467 

chrisdennis:


chrisdennis:
I believe this falls under "contributions which do no require a full Contributor's License Agreement", as described here: http://www.quartz-scheduler.org/community/contribute.html#contributions-which-do-not-require-full-contributors-license-agreement-cla

@jgallimore If you could confirm that you agree with a simple yes below that would be great!

jgallimore:
Yes, I agree. Thank you!

PR #509, opened 2019-10-15
Fixes #508

### LINKED ISSUE #508: Error with H2 1.4.200
Starting with H2 1.4.200 recently released, the following exception is raised:

```
org.quartz.JobPersistenceException: Couldn't remove trigger: Table "QRTZ_SIMPROP_TRIGGERS" not found; SQL statement:
DELETE FROM QRTZ_SIMPROP_TRIGGERS WHERE SCHED_NAME = 'ClusteredScheduler' AND TRIGGER_NAME = ? AND TRIGGER_GROUP = ? [42102-200] {DELETE FROM QRTZ_SIMPROP_TRIGGERS WHERE SCHED_NAME = 'ClusteredScheduler' AND TRIGGER_NAME = ? AND TRIGGER_GROUP = ?} [code=42102, state=42S02]
```

Works until H2 1.4.199

ilgrosso:


quartz-core/src/main/resources/org/quartz/impl/jdbcjobstore/tables_h2.sql
```diff
@@ -77,8 +77,7 @@ CREATE TABLE QRTZ_SIMPLE_TRIGGERS (
   TIMES_TRIGGERED BIGINT NOT NULL
 );
 
-CREATE TABLE qrtz_simprop_triggers
-  (          
+CREATE TABLE QRTZ_SIMPROP_TRIGGERS (
```
chrisdennis: I understand the desire to align casing across the file, but could we keep the bug fix to just the single character change necessary to fix the bug.

ilgrosso: I am not completely sure that the different case between table definition and queries is not part of the issue with latest H2 1.4.200.
Hence, to stay safe, I changed that too.

katzyn: Unquoted names are case insensitive in H2 by default, but they are case-sensitive with some non-default settings. It's safer to use the same case of letters everywhere.

chrisdennis: Okay, I'll bow to the peer pressure.

chrisdennis:
@ilgrosso are you interested in rebasing this to be a daggy PR, or are you prepared to wait for me to do that?

chrisdennis:
Also are you happy that this change be considered a trivial one and that therefore no contributor agreement is required? (http://www.quartz-scheduler.org/documentation/2.4.0-SNAPSHOT/contribute.html#contributions-which-do-not-require-full-contributors-license-agreement-cla)

P.S .It would also be good to get a more descriptive commit message.

ilgrosso:
> @ilgrosso are you interested in rebasing this to be a daggy PR, or are you prepared to wait for me to do that?

@chrisdennis What is a "daggy PR"?

> Also are you happy that this change be considered a trivial one and that therefore no contributor agreement is required?

I don't mind at all if my name does not appear among contributors; signing a CLA would be not an issue, anyway.

> P.S .It would also be good to get a more descriptive commit message.

Ok, commit message changed, hope it is fine now.

Any timing about when next release including this fix will be available? AFAICT there is at least spring-projects/spring-boot#18593 waiting for it.

chrisdennis:
A daggy PR would be one like this: https://wiki.monotone.ca/DaggyFixes/

Essentially you make the commit fix a direct child of the bug that introduced the commit. This mean the fix can then be merged to any other child of the buggy commit.  This way all fixed versions share the same fixing commit.  In this case that commit is likely the initial commit in the repo since this bug is I suspect older than this git repo itself. I'll cherry-pick your fix commit, rebase appropriately and then create a new PR.

Release should be soon... this week barring any further complications. Possibly even today.

chrisdennis:
New PR is #513

ilgrosso:
Thanks 👍

Issue #189, opened 2017-10-17
NPE thrown when acquiring next trigger due to null next fire time value


Dwisf:
During a soak testing of our application using Quartz (version 2.2.3), we notice the following exceptions being thrown from time to time: 

```
org.quartz.core.ErrorLogger ERROR --- An error occurred while scanning for the next triggers to fire. 
java.lang.NullPointerException: null
	at org.quartz.impl.jdbcjobstore.JobStoreSupport.acquireNextTrigger(JobStoreSupport.java:2837)
	... 5 common frames omitted
Wrapped by: org.quartz.JobPersistenceException: Couldn't acquire next trigger: null
	at org.quartz.impl.jdbcjobstore.JobStoreSupport.acquireNextTrigger(JobStoreSupport.java:2864)
	at org.quartz.impl.jdbcjobstore.JobStoreSupport$40.execute(JobStoreSupport.java:2759)
	at org.quartz.impl.jdbcjobstore.JobStoreSupport$40.execute(JobStoreSupport.java:2757)
	at org.quartz.impl.jdbcjobstore.JobStoreSupport.executeInNonManagedTXLock(JobStoreSupport.java:3803)
	at org.quartz.impl.jdbcjobstore.JobStoreSupport.acquireNextTriggers(JobStoreSupport.java:2756)
	at org.quartz.core.QuartzSchedulerThread.run(QuartzSchedulerThread.java:272)
```
Line 2837 of `JobStoreSupport.java` corresponds to the following code: 

```
...
if (nextTrigger.getNextFireTime().getTime() > batchEnd) {
...
```

That is `getNextFireTime()` is null. The API for this method states `...If the trigger will not fire again, null will be returned`. It looks like this code needs to check if `getNextFireTime()` is null before doing a `getTime()`.

Is this a correct assessment?

bsiamionau:
Any progress on this?

sudhanmoghe:
We are also getting this error. 
We started getting this error after adding 2nd node to cluster. We never got this error with single node. 
We see the exception stack trace in logs. 285 in last one month. But, no issue in actual job execution. No failed or pending job. 
We only have CRON jobs or Simple JOBS. I don't see any NEXT_FIRE_TIME < 0 in TRIGGERS table. Not sure what is causing the NPE and how to fix that. 
We are running cluster with this config.
org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreTX
[...]

PR #713, opened 2021-08-04
#580 fix wrong warning from #189 change

### LINKED ISSUE #580: Wrong warning log message due to NPE fix
The [fix](https://github.com/quartz-scheduler/quartz/commit/801abe0f67e80a88609804ad78dcaee3402c199b) of #189 leads to wrong warning log messages on clustered quartz instances with cron triggers. What happens:

- instance 1 aquires some triggers
- while processing, CronTriggerImpl#getFireTimeAfter nulls nextFireTime and the trigger is persisted
- meanwhile instance 2 also aquires some trigger, including the one before
- because nextFireTime is null the warning is logged
- BUT: this trigger couldn't have been aquired some lines later in insance 2 as instance 1 already changed STATE_WAITING to  STATE_ACQUIRED

I don't see an easy fix in [JobStoreSupport](https://github.com/quartz-scheduler/quartz/blob/v2.3.2/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java#L2883) because if batchEnd is checked nextFireTime must not be null, but only when the job is aquired the warning message could have made sense. So I see two posibilities:

1. Remove the new log message and check for null. It doesn't matter that batchEnd check is skipped on a non-aquirable trigger:
`if (nextFireTime != null && nextFireTime.getTime() > batchEnd) {`

2. Change the log level and text (and use getLog()):
`getLog().info("Trigger {} returned null on nextFireTime and yet still exists in DB. This should only happen in clustered mode", nextTrigger.getKey());`

Munich81:
Warning regularly appears due to race condition on multi-instance environments although nothing is wrong

In submitting this contribution, I agree to the current Software AG contributor agreement as referred to here: https://github.com/quartz-scheduler/contributing/CONTRIBUTING.md
Err ... Not Found?

This PR...
## Changes
- Fixes a wrong warning message

## Checklist
- [x] tested locally
- [/] updated the docs - Does not seem necessary
- [/] added appropriate test - Does not seem necessary / No existing test
- [x] signed-off on the above mentioned SoftwareAG contributor agreement via -s on my commits. 
  (You can use a [browser extension](https://github.com/scottrigby/dco-gh-ui) if you're not using command-line)

Fixes #580

### DIFF
--- a/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
+++ b/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
@@ -2850,8 +2850,13 @@ public abstract class JobStoreSupport implements JobStore, Constants {
                 long batchEnd = noLaterThan;
 
                 for(TriggerKey triggerKey: keys) {
-                    // If our trigger is no longer available, try a new one.
-                    OperableTrigger nextTrigger = retrieveTrigger(conn, triggerKey);
+                    // If our trigger is no longer available or no longer waiting, try a new one.
+                    OperableTrigger nextTrigger;
+                    try {
+                        nextTrigger = getDelegate().selectTriggerInState(conn, triggerKey, STATE_WAITING);
+                    } catch (Exception e) {
+                        throw new JobPersistenceException("Couldn't retrieve next trigger: " + e.getMessage(), e);
+                    }
                     if(nextTrigger == null) {
                         continue; // next trigger
                     }
--- a/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/DriverDelegate.java
+++ b/quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/DriverDelegate.java
@@ -640,6 +640,7 @@ public interface DriverDelegate {
      */
     List<OperableTrigger> selectTriggersForCalendar(Connection conn, String calName)
         throws SQLException, ClassNotFoundException, IOException, JobPersistenceException;
+
     /**
      * <p>
      * Select a trigger.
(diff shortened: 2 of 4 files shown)


quartz-core/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
```diff
@@ -2902,6 +2891,18 @@ public Boolean validate(Connection conn, List<OperableTrigger> result) throws Jo
                     if (rowsUpdated <= 0) {
                         continue; // next trigger
                     }
+
```
vionixt, line 2894: Should we update the trigger state, if nextFireTime is null?

Munich81: Well it should not be in a state that it will be updated ... but you are right.

Thinking about the problem, when nextFireTime is null already the state certainly has changed as well. I think the best approach would be to check whether nextTrigger still is STATE_WAITING.

What do you think? I would be prefer to create a new selectTriggerInState function than make an additional query with selectTriggerState?

Munich81:
Changed the fix to include state when fetching trigger (and not update the trigger)

jhouserizer:
Contributing link in PR template corrected.  Sorry, and thanks for pointing it out.   Proper link is: https://github.com/quartz-scheduler/contributing/blob/main/CONTRIBUTING.md

jhouserizer:
@Munich81  - for clarity, is it the "Trigger {} returned null on nextFireTime and yet still exists in DB!" warning message that you're referring to?

Munich81:
> @Munich81 - for clarity, is it the "Trigger {} returned null on nextFireTime and yet still exists in DB!" warning message that you're referring to?

Yes.

(forgot sign-off in first amend, added with 09f5fa9)

PR #723, opened 2021-08-30
Support for building on java11

konopka:

This PR...
## Changes
- fix compilation errors on JDK11
- added dependencies for modules which were removed from JDK11
- fix javadoc issues

## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test

Fixes #

### DIFF
--- a/quartz-jobs/pom.xml
+++ b/quartz-jobs/pom.xml
@@ -85,7 +85,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
-        <version>2.9</version>
+        <version>3.3.0</version>
         <configuration>
           <skip>${skipJavadoc}</skip>
           <quiet>true</quiet>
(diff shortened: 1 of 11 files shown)


quartz-jobs/pom.xml
```diff
@@ -73,14 +73,19 @@
       <version>2.7</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish.corba</groupId>
+      <artifactId>rmic</artifactId>
+      <version>4.2.4</version>
```
melloware, line 79: Shouldn't this be `<scope>provided</scope>` so its not enforced as a transitive compile time dependency?  The rest of the JARS in this pom are all provided or test.

melloware (changes requested):

quartz-jobs/pom.xml
konopka: Yes it should, updated now

melloware (approved):

melloware:
What we do is build two versions of our library PrimeFaces. 1 for javax and 1 for jakarta. This can be done with a Maven plugin so it doesn't require two branches of code.

PR #725, opened 2021-09-07
Stale plugin config

mathieucarbou:
Signed-off-by: Mathieu Carbou <mathieu.carbou@gmail.com>

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,2 +1,58 @@
+# Configuration for probot-stale - https://github.com/probot/stale
 
-# Empty (using defaults from https://github.com/probot/stale)
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - todo
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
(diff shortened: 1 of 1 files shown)


akomakom:
@mathieucarbou I guess you want it to be more aggressive than https://github.com/probot/.github/blob/master/.github/stale.yml ?

mathieucarbou:
> @mathieucarbou I guess you want it to be more aggressive than https://github.com/probot/.github/blob/master/.github/stale.yml ?

The default values are 60 / 7 (https://github.com/probot/stale) which are the values we were using according to their doc.

The link you are pointing to is their own configuration for their own repositories: they decided to go with 3 months.

Whatever the value we pick I am fine with it, but this PR doesn't change the previous delay.

.github/stale.yml
```diff
@@ -1,2 +1,59 @@
...
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - todo
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
```
mathieucarbou, line 21: changed here

.github/stale.yml
```diff
@@ -1,2 +1,59 @@
...
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - todo
```
mathieucarbou, line 15: changed here

.github/stale.yml
```diff
@@ -1,2 +1,59 @@
...
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: stale
```
mathieucarbou, line 27: changed here (and renamed the current label wontfix to stale)

akomakom:
@mathieucarbou what's your opinion about overriding all (as you seem to be doing) vs overriding only settings different from defaults?

mathieucarbou:
> @mathieucarbou what's your opinion about overriding all (as you seem to be doing) vs overriding only settings different from defaults?

My opinion for this use case is that I prefer to have them all defined (so that they never changed without us doing so) and also to specifically ensure we have the values we want. Because before, it wasn't clear either whether we were at 60 days or 90 days ;-)

akomakom (approved):

PR #745, opened 2021-12-02
Fix the wrong parameter used in createScheduler

topikachu:

This PR Fix the wrong parameter used in createScheduler

## Changes
- Fix the wrong parameter used in createScheduler

## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test

Fixes #


topikachu:
It seems the error is caused by the wrong pipeline configuration.

PR #753, opened 2021-12-16
Added missing indexes for SQL Server.

0x7d7b:
SQL Server does not automatically create indexes when using foreign key constraints. This PR adds them.


0x7d7b:
Still relevant.

jhouserizer:
Thanks 0x7d7b .   

Note (to team) before merging and closing (and forgetting about) this PR: We'll probably want this to go into other branches (e.g. 2.3.x) as well.

lahma:
I think the fields are now indexed twice. There are already clustered indexes forming the primary key:

```sql
ALTER TABLE [dbo].[QRTZ_CRON_TRIGGERS] WITH NOCHECK ADD
  CONSTRAINT [PK_QRTZ_CRON_TRIGGERS] PRIMARY KEY  CLUSTERED
  (
    [SCHED_NAME],
    [TRIGGER_NAME],
    [TRIGGER_GROUP]
  )  ON [PRIMARY]
GO
```

These extra indexes would make sense if there would be some generated primary key I guess.

Issue #744, opened 2021-11-14
Problems starting spring boot application:Unique-Constraint »qrtz_locks_pkey«


somera:
Hi,

sometimes I can't start one of my instances (start parallel on 4 computers) cause I get the Unique-Constraint »qrtz_locks_pkey« error

```
71956 [main] ERROR o.s.boot.SpringApplication - Application run failed
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'quartzScheduler' defined in class path resource [org/springframework/boot/autoconfigure/quartz/QuartzAutoConfiguration.class]: Invocation of init method failed; nested exception is org.quartz.impl.jdbcjobstore.LockException: Failure obtaining db row lock: FEHLER: aktuelle Transaktion wurde abgebrochen, Befehle werden bis zum Ende der Transaktion ignoriert [See nested exception: org.postgresql.util.PSQLException: FEHLER: aktuelle Transaktion wurde abgebrochen, Befehle werden bis zum Ende der Transaktion ignoriert]
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1799)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:594)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:516)
        at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:324)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:322)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
        at org.springframework.context.support.DefaultLifecycleProcessor.getLifecycleBeans(DefaultLifecycleProcessor.java:289)
        at org.springframework.context.support.DefaultLifecycleProcessor.startBeans(DefaultLifecycleProcessor.java:141)
        at org.springframework.context.support.DefaultLifecycleProcessor.onRefresh(DefaultLifecycleProcessor.java:122)
[...]

PR #763, opened 2022-02-01
Fix StdRowLockSemaphore retry mechanism

FranciscoCagigal:

This PR fixes the retry mechanism of StdRowLockSemaphore.

## Changes
- If connection is never rollbacked upon error and auto-commit is disabled, the transaction only ends when commit is called. In this case the retry never succeeds if it fails the first time when acquiring the lock.

## Checklist
- [x] tested locally

Fixes #

I believe it might be related to this already open issued - https://github.com/quartz-scheduler/quartz/issues/744.


FranciscoCagigal:
@akomakom if you have some time :p

jhouserizer:
Looks good, thanks @FranciscoCagigal  .

Note to team: we'll want this fix in branches (e.g. 2.3.x) as well.

PR #772, opened 2022-02-28
Fix n plus one problem in scheduler

konopka:

This PR...
## Changes
Fix N+1 problem in scheduler
Load triggers with single query instead of doing
1. load trigger keys to acquire
2. load trigger for each trigger key in a loop
3. load job for each trigger in a loop

## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test

Fixes #


konopka:
Yes, relevant

PR #774, opened 2022-03-01
Interrupt the scheduler thread when shutdown is occurring, in case it …

jhouserizer:

This PR...
## Changes
- Interrupt the scheduler thread when shutdown is occurring, in case it is blocked in a re-try loop on acquiring/releasing triggers.

## Checklist
- [X] tested locally
- [X] signed-off on the above mentioned SoftwareAG contributor agreement via `git commit -s` on my commits. 

Fixes #

### DIFF
--- a/quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
+++ b/quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
@@ -166,6 +166,7 @@ public class QuartzSchedulerThread extends Thread {
     void halt(boolean wait) {
         synchronized (sigLock) {
             halted.set(true);
+            this.interrupt();
 
             if (paused) {
                 sigLock.notifyAll();
@@ -276,6 +277,11 @@ public class QuartzSchedulerThread extends Thread {
                 }
 
                 int availThreadCount = qsRsrcs.getThreadPool().blockForAvailableThreads();
+                synchronized (sigLock) {
+                    if (halted.get()) {
+                        break;
+                    }
+                }
                 if(availThreadCount > 0) { // will always be true, due to semantics of blockForAvailableThreads...
 
                     List<OperableTrigger> triggers;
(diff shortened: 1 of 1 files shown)


quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
```diff
@@ -276,6 +277,11 @@ public void run() {
                 }
 
                 int availThreadCount = qsRsrcs.getThreadPool().blockForAvailableThreads();
+                synchronized (sigLock) {
```
prasantakumar82, line 280: **_halted_** is an atomic variable, is synchronization required to fetch it's value?

prasantakumar82: though synchronization is not required, but no harm by keeping it

quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
```diff
@@ -166,6 +166,7 @@ void togglePause(boolean pause) {
     void halt(boolean wait) {
         synchronized (sigLock) {
             halted.set(true);
+            this.interrupt();
```
prasantakumar82, line 169: Why can't we delay this? I mean just before invoking join() method

quartz-core/src/main/java/org/quartz/core/QuartzSchedulerThread.java
jhouserizer: > **_halted_** is an atomic variable, is synchronization required to fetch it's value?

I agree - I did it to remain consistent with existing code.   I think that long ago this may have not been an AtomicBoolean - because the code even pre-dates java 1.5 (when Atomic types were introduced) , perhaps it wasn't fully cleaned up when switching to atomic, or perhaps it is still purposely trying to align with all the other points that sync up on scheduling changes.    

So to repeat: Overall I believe it unnecessary, but again, for consistency...

prasantakumar82 (approved):

PR #802, opened 2022-05-19
Implement CronExpression.getTimeBefore

srgrr:
Use the fact that if for some cron expression and some instant t the smallest valid firing time greater than t is v then for some t' > t the smallest firing time greater than t' will be some v' > v to binary search in the interval [epoch, endTime).

This idea allows us to find the closest past instant that would have been a valid firing time very quickly.

Signed-off-by: Sergio Rodriguez <sergio.rodriguez@streamsets.com>

This PR...
## Changes
- Implement the CronExpression.getTimeBefore method, which allows to find the closest instant strictly before another given instant that would have made the CronExpression to trigger.

## Checklist
- [X] tested locally
- [ ] updated the docs
- [X] added appropriate test
- [X] signed-off on the above mentioned SoftwareAG contributor agreement via `git commit -s` on my commits. 

Fixes #

### DIFF
--- a/quartz-core/src/main/java/org/quartz/CronExpression.java
+++ b/quartz-core/src/main/java/org/quartz/CronExpression.java
@@ -18,7 +18,10 @@
 package org.quartz;
 
 import java.io.Serializable;
+import java.text.DateFormat;
 import java.text.ParseException;
+import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


srgrr:
Not sure about what docs affect this implementation.

Issue #903, opened 2023-04-17
Convert to gradle build, update dependencies, require java 8


jhouserizer:
Overhaul  bit-rotted, build system
Update to support of more recent versions of 3rd party libraries
Update to require at least Java 8 (used to be java 7)

jhouserizer:
Resolved with #904

PR #904, opened 2023-04-17
Restructure project to better match release artifacts, use gradle #903, #904

jhouserizer:
Also:

* rework project modules
* revamp examples to be executable by gradle
* remove usage of rmic
* remove TerracottaJobStore
*


azure-pipelines.yml
```diff
@@ -26,11 +26,16 @@ resources:
 jobs:
   - template: build-templates/maven-common.yml@templates
     parameters:
-      jobName: Java11
-      jdkVersion: '1.11'
+      jobName: Java8
+      jdkVersion: '1.1_8'
```
akomakom, line 30: Looks wrong

azure-pipelines.yml
```diff
@@ -26,11 +26,16 @@ resources:
 jobs:
   - template: build-templates/maven-common.yml@templates
```
chrisdennis, line 27: This needs switching to a grade template to match the new build system.

build.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
```
chrisdennis, line 1: Top level project doesn't need any plugins applied.

build.gradle
```diff
@@ -0,0 +1,16 @@
...
+
+subprojects {
+    apply plugin: 'java-library'
+}
+
+allprojects {
+    group 'org.quartz-scheduler'
+    version '2.4.0-SNAPSHOT'
+
+    java.sourceCompatibility(JavaVersion.VERSION_11)
```
chrisdennis, line 14: This should probably either be done with toolchains, or with a release setting on the relevant compiler tasks: https://docs.gradle.org/current/userguide/toolchains.html

examples/build.gradle
```diff
@@ -0,0 +1,121 @@
+plugins {
+    id 'java'
+}
+
+
+repositories {
```
chrisdennis, line 6: If we're going to have a subprojects block in the top level build.gradle then we may as well stick this there.

examples/build.gradle
```diff
@@ -0,0 +1,121 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    implementation project(path: ":quartz")
```
chrisdennis, line 11: No need for the `path: ":quartz'` you can just call `project(':quartz')`.

examples/build.gradle
```diff
@@ -0,0 +1,121 @@
...
+    implementation project(path: ":quartz-jobs")
+
+    implementation "org.slf4j:slf4j-api:$slf4jVersion"
+    implementation "org.slf4j:slf4j-log4j12:$slf4jVersion"
```
chrisdennis, line 15: Do we really depend on a logging provider at compile time? Should this be `runtimeOnly` scoped?

examples/build.gradle
```diff
@@ -0,0 +1,121 @@
...
+}
+
+
+task runExample1(type: Exec) {
```
chrisdennis, line 19: There's a `JavaExec` task type for exactly this. Better syntax here would be something like:
```groovy
tasks.withType(JavaExec).configureEach() {
  classpath = sourcsets.main.runtimeClasspath
  group = '...'
}

tasks.register('runExample1', JavaExec) {
  description = '...'
  mainClass = 'org.quartz.examples.example1.SimpleExample'
}
```

quartz/build.gradle
```diff
@@ -0,0 +1,94 @@
...
+}
+
+repositories {
+    mavenCentral()
+}
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId = "$project.group"
```
chrisdennis, line 12: These are all the defaults so no need to set them here (from memory).

chrisdennis (changes requested):

build.gradle
jhouserizer: actually I tried removing them from here and things broke.  I guess I'll fool around with it some more.

PR #906, opened 2023-04-19
Add Nexus Publish plugin and signing

akomakom:
* moves duplicated logic to root project
* adds signing
* adds pom.xml and pom.properties to the jar
* add developers section to pom
* turns of `module.json` publishing


jhouserizer:
dco / sign-off not required for akom (SAG employee).

PR #925, opened 2023-06-02
Add a sql LIMIT clause for acquire next triggers

weih-kahoot:
In the case where database has a lot of soon-to-be-fired-jobs, the query can be slow as it needs to do extra sort by next_fire_time, priority, the sql LIMIT clause should enable database to be more efficient.

This PR...
## Changes
-

## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test

Fixes #

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/PostgreSQLDelegate.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/PostgreSQLDelegate.java
@@ -44,6 +44,17 @@ import org.slf4j.Logger;
  */
 public class PostgreSQLDelegate extends StdJDBCDelegate {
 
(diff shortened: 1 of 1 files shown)


lahma:
Not all databases support LIMIT, MS SQL Server for example. 

In .NET version we override in delegate and add suitable extra: https://github.com/quartznet/quartznet/blob/28b5faec68aa7f3227b3e744780617ada0f0eb6f/src/Quartz/Impl/AdoJobStore/SqlServerDelegate.cs#L37

weih-kahoot:
@lahma Thanks! I've no knowledge of MS SQL Server, I've updated PR to only apply LIMIT for Postgres

weih-kahoot:
@jhouserizer can you check if this change makes sense when you have some time?

PR #1082, opened 2023-11-23
fix tutorial link for github navigation

dcolazin:
This change is a minimal fix on documentation; in any case, if needed, I will agree to the current Software AG contributor agreement as referred to here: 

This PR fixes the link between the markdown/asciidoc tutorial files.

## Checklist
- [x] updated the docs


dcolazin:
abandoned, this will break the website; a way to support both github links and the website should be found

PR #1086, opened 2023-12-08
When storing job with replace check update count returned

### LINKED ISSUE #1085: Foreign key constraint violation when storing trigger to non-durable job on MariaDB/MySQL
When scheduling a trigger for a non-durable job it may happen that the foreign key of trigger to job is violated on MariaDB/MySQL. This can happen if another trigger for the job completes concurrently while storing the job with the replace option set.

MariaDB/MySQL implement _consistent nonlocking reads_ and _multi-versioned concurrency control_, see <https://dev.mysql.com/doc/refman/8.0/en/innodb-consistent-read.html>. The `JobStoreSupport.storeJob(Connection, JobDetail, boolean)` method uses an if-exists-update-else-insert approach that in case above exhibits a race specific to the way InnoDB is implemented. Here is what happens:

1. The `SELECT` of `jobExists` reads the existing job row from the database. This is a _nonlocking read_.
2. Now, job execution completes in a concurrent transaction. The trigger of that execution gets deleted and since it is the last one referencing the non-durable job, the job is deleted as well. This concurrent operation commits.
3. The operation of step 1 continues, the `updateJobDetail` issues an `UPDATE` statement. Because of _multi-versioned concurrency control_ the delete of step 2 becomes visible, and the update does not hit any row.
4. Next, a trigger for the job is inserted which results in the foreign key constraint violation because the job does not exist.

I was above to reproduce the above more or less reliably in integration test of a larger software system we develop. The race condition hits us in 1 out of 100 `scheduleJob(jobDetail, trigger, true)` operations. I'll add a PR fixing this by checking the returned row count of the update statement shortly.

t-beckmann:
This is to detect non-durable job suddenly disappearing when concurrently completing. On MariaDB/MySQL the fall-through code path select-update-insert can be taken due to consistent nonblocking read and multi-version concurrency control.

## Changes
The code change checks the update count returned when storing a job.

## Checklist
- [x] tested locally
- [x] updated the docs (not applicable, bug fix only)
- [x] added appropriate test (not applicable, storeJob method covered in tests already)

Fixes #1085

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
@@ -1112,9 +1112,12 @@ public abstract class JobStoreSupport implements JobStore, Constants {
                 if (!replaceExisting) { 
                     throw new ObjectAlreadyExistsException(newJob); 
                 }
-                getDelegate().updateJobDetail(conn, newJob);
-            } else {
-                getDelegate().insertJobDetail(conn, newJob);
+                if (getDelegate().updateJobDetail(conn, newJob) > 0) {
+                    return;
+                }
+            }
+            if (getDelegate().insertJobDetail(conn, newJob) < 1) {
+                throw new JobPersistenceException("Couldn't store job. Insert failed.");
             }
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


jhouserizer:

PR #1094, opened 2024-01-06
use double-checked locking to obtain delegate

hwj0123:

This PR...
## Changes
- In org.quartz.impl.jdbcjobstore.JobStoreSupport，use double-checked locking to obtain delegate 

## Checklist
- [ ] tested locally
- [ ] updated the docs
- [ ] added appropriate test
- [ ] signed-off on the above mentioned SoftwareAG contributor agreement via `git commit -s` on my commits. 

Fixes #

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
@@ -111,7 +111,7 @@ public abstract class JobStoreSupport implements JobStore, Constants {
 
     protected HashMap<String, Calendar> calendarCache = new HashMap<String, Calendar>();
 
-    private DriverDelegate delegate;
+    private volatile DriverDelegate delegate;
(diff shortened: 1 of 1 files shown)


hwj0123:
Hello, I saw that the org.quartz.impl.jdbcjobstore.JobStoreSupport#getDelegate method uses synchronized to obtain the DriverDelegate. I wonder if it would be better to use a double-check lock here. I look forward to your reply. Thank you.

ShivamNagpal:
These might be good reads

- https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html#:~:text=The%20fix%20to%20this%20is,functions%20and%20members...%20%7D
- https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2168/

The Sonar Rule for the Double Checked Locking is tagged as a Blocker issue. I wouldn't recommend to use this.

jhouserizer:

PR #1148, opened 2024-08-14
Update CronExpression.java

KPDeweese:
Update error messaging to match corresponding error conditions in the increment range validation logic

### DIFF
--- a/quartz/src/main/java/org/quartz/CronExpression.java
+++ b/quartz/src/main/java/org/quartz/CronExpression.java
@@ -709,9 +709,9 @@ public final class CronExpression implements Serializable, Cloneable {
 
     private void checkIncrementRange(int incr, int type, int idxPos) throws ParseException {
         if (incr > 59 && (type == SECOND || type == MINUTE)) {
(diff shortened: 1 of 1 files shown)


Nishant9771 (approved):
### It looks like you have some redundant throw new ParseException statements in your code. You only need one exception for each condition.


private void checkIncrementRange(int incr, int type, int idxPos) throws ParseException {
    if (incr > 59 && (type == SECOND || type == MINUTE)) {
        throw new ParseException("Increment > 59 : " + incr, idxPos);
    } else if (incr > 23 && (type == HOUR)) {
        throw new ParseException("Increment > 23 : " + incr, idxPos);
    } else if (incr > 31 && (type == DAY_OF_MONTH)) {
        throw new ParseException("Increment > 31 : " + incr, idxPos);
    } else if (incr > 7 && (type == DAY_OF_WEEK)) {
        throw new ParseException("Increment > 7 : " + incr, idxPos);
    }
}

jhouserizer:

PR #1153, opened 2024-09-02
fix: use ClassLoaderObjectInputStream to resolve class loading issue …

srmsh:
…in getObjectFromBlob

Quartz currently uses ObjectInputStream in getObjectFromBlob, which can lead to class loading issues when the class being deserialized is not loaded by the same class loader as Quartz. This commit replaces ObjectInputStream with ClassLoaderObjectInputStream to ensure that the correct class loader is used during deserialization.

- Replaced ObjectInputStream with ClassLoaderObjectInputStream in getObjectFromBlob method.
- Added a new ClassLoaderObjectInputStream class to handle custom class loading.

This change resolves the issue where deserialized objects are not recognized due to class loader mismatches, improving compatibility in environments with multiple class loaders.

This PR...
## Checklist

Fixes #


jhouserizer:

PR #1190, opened 2024-10-14
Refactoring code and apply some language migration

bcode2:
This PR...

Fixes issue #

## Changes
- fix unnecessary String conversion
- fix redundant variable declaration
- Replace try-catch-finally by try-with-resources
- replace loop by enhanced for loop
- simplify boolean expression
- use StringBuilder instead of String concatenation
- replace concatenation with parameterized log message
- remove unused imports
- remove unnecessary semi-colon
- fix unnecessary call to toString() method
- Language migration.Replace explicit type by diamond operator
- set certain fields as final
- replace several if-else sentences by switch statements to gain readability
- use isEmpty() to check empty collections

-----------------
## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test
-----------------

### DIFF
--- a/quartz/src/main/java/org/quartz/CronExpression.java
+++ b/quartz/src/main/java/org/quartz/CronExpression.java
@@ -1204,7 +1204,7 @@ public final class CronExpression implements Serializable, Cloneable {
 
             // get second.................................................
             st = seconds.tailSet(sec);
-            if (st != null && st.size() != 0) {
+            if (st != null && !st.isEmpty()) {
                 sec = st.first();
             } else {
                 sec = seconds.first();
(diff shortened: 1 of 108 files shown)


melloware (approved):

bcode2:
> Good stuff, I'll try to make time soon to review. Would you be willing to port this into the 2.5.0 line as well, or should I factor-in that I will need to make time for that?

in fact all those changes apply to 2.5 too. I have a lot more fixes to apply to both branches, but as in production I am using 2.4-rc, i prefered to apply my first commit   to 2.4,but don't worry I will apply the same fixes to 2.5 ,it is relatively easy

jhouserizer (approved):

jhouserizer:
I reviewed the changes and they all look safe.

These are nice modernizations/cleanup, but this sort of change is also somewhat like gratuitous reformatting of code - in that is causes a mass set of changes that don't affect logic/behavior, but does create a painful boundary in revisions across which diffs (on actual logic) are very difficult.   As such if there are to be any other similar sets of changes, the sooner the better, so as not to create multiple diffing boundaries between revisions.

PR #1206, opened 2024-10-16
Refactor branch 2.5, clean code

bcode2:
This PR...

Fixes issue #

## Changes
refactor: replace while loop with enhanced for loop
refactor: language migration; simplify maps operations
style: improve visibility change l by L for long literals
refactor: replace addAll() with  parameterized constructor
refactor: set serialVersionUID  fields as static and final
refactor: language migration, remove unnecessary boxing
refactor: language migration, use lambda expression to simplify code and remove unused imports
refactor: remove redundant variable declaration
performance: language migration
refactor: simplify boolean expression

-----------------
## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test
- [x] signed-off on the DCO referenced in the CONTRIBUTING link below via `git commit -s` on my commits, and submit this code under terms of the Apache 2.0 license and assign copyright to the Quartz project owners
  (If you're not using command-line, you can use a [browser extension](https://github.com/scottrigby/dco-gh-ui) )
-----------------
In submitting this contribution, I agree to the terms of contributing as referred to here: 
https://github.com/quartz-scheduler/contributing/blob/main/CONTRIBUTING.md

PR #1210, opened 2024-10-16
Fix #1209 redundant check

### LINKED ISSUE #1209: Redundant check
https://github.com/quartz-scheduler/quartz/blob/199c8f1111f2034f208300a8f9b372a9f37cf74f/quartz/src/main/java/org/quartz/impl/triggers/DailyTimeIntervalTriggerImpl.java#L875-L882

You have two checks that verify whether daysOfWeek is null or empty, but the second check (else if(daysOfWeek.isEmpty())) is redundant because you've already checked for the same condition in the previous line.

melloware:
Fix #1209 redundant check


jhouserizer:
Redundant fix with #1206 ?

PR #1221, opened 2024-10-18
Fix #1161 Add Oracle23 SQL script using BOOLEAN

### LINKED ISSUE #1161: [Oracle 23] VARCHAR2(1) breaking change
**Current behavior**

Job can no more be registered inside job detail table.

```
Caused by: org.quartz.JobPersistenceException: Couldn't store job: ORA-12899: valeur trop grande pour la colonne "QUARKUS"."QRTZ_JOB_DETAILS"."IS_DURABLE" (réelle : 5, maximum : 1)
```

**Expected behavior**

Job should be stored and next executed as expected.

**Additional information**

Oracle 23 has introduced a breaking change regarding the way a boolean is represented.
Before Oracle 23 a boolean used a **VARCHAR2(1)** and this way has been supported from many years by the oracle jdbc driver.

Now, Oracle 23 has introduced a new type **BOOLEAN** to represent a boolean type.
But, the driver is no more compatible - except if an additional property is used to deactivate the new way to handle boolean type.

Full story here with a reproducer : https://github.com/quarkusio/quarkus/issues/43720

**How to fix**

Provide an update script for Oracle database 23 and plus to update boolean columns **VARCHAR2(1)** to **BOOLEAN**
or
provide a new full initialization tables for Oracle database 23 and plus.

Multiples tables are impacted.

Regards,
Damien

melloware:
Fix #1161 Add Oracle23 SQL script using BOOLEAN

Oracle 23 uses BOOLEAN instead of VARCHAR2(1)

### DIFF
--- a/quartz/src/main/resources/org/quartz/impl/jdbcjobstore/tables_oracle23.sql
+++ b/quartz/src/main/resources/org/quartz/impl/jdbcjobstore/tables_oracle23.sql
@@ -1,0 +1,191 @@
+--
+-- A hint submitted by a user: Oracle DB MUST be created as "shared" and the 
+-- job_queue_processes parameter  must be greater than 2
+-- However, these settings are pretty much standard after any
+-- Oracle install, so most users need not worry about this.
+--
+-- Many other users (including the primary author of Quartz) have had success
+-- running in dedicated mode, so only consider the above as a hint ;-)
+--
+
+delete from qrtz_fired_triggers;
+delete from qrtz_simple_triggers;
+delete from qrtz_simprop_triggers;
+delete from qrtz_cron_triggers;
+delete from qrtz_blob_triggers;
+delete from qrtz_triggers;
+delete from qrtz_job_details;
+delete from qrtz_calendars;
+delete from qrtz_paused_trigger_grps;
+delete from qrtz_locks;
+delete from qrtz_scheduler_state;
+
+drop table qrtz_calendars;
+drop table qrtz_fired_triggers;
+drop table qrtz_blob_triggers;
+drop table qrtz_cron_triggers;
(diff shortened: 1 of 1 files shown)


bcode2:
Should we worry about this?

`
    /**
     * Retrieves the value of the designated column in the current row
     * of this <code>ResultSet</code> object as
     * a <code>boolean</code> in the Java programming language.
     *
     * <P>If the designated column has a datatype of CHAR or VARCHAR
     * and contains a "0" or has a datatype of BIT, TINYINT, SMALLINT, INTEGER or BIGINT
     * and contains  a 0, a value of <code>false</code> is returned.  If the designated column has a datatype
     * of CHAR or VARCHAR
     * and contains a "1" or has a datatype of BIT, TINYINT, SMALLINT, INTEGER or BIGINT
     * and contains  a 1, a value of <code>true</code> is returned.
     *
     * @param columnLabel the label for the column specified with the SQL AS clause.  If the SQL AS clause was not specified, then the label is the name of the column
     * @return the column value; if the value is SQL <code>NULL</code>, the
     * value returned is <code>false</code>
     * @exception SQLException if the columnLabel is not valid;
     * if a database access error occurs or this method is
     *            called on a closed result set
     */
    boolean getBoolean(String columnLabel) throws SQLException;`

melloware:
Good question.  We have this discussion going on https://github.com/quarkusio/quarkus/issues/43720#issuecomment-2399045448

And I have been running fine with these changes as well so i think its all working

bcode2:
Perfect!!!!!! @melloware  :+1:

jhouserizer (approved):

jhouserizer:
Thanks!

Issue #1101, opened 2024-02-13
Quartz Extension not working with Micosoft SQL server since 3.2.x


chvsnraju:
We use Quartz scheduler for our cronjob executions. We started with Quarkus 2.6.x, and it is working until 3.1.x. When we migrated to 3.2.x, we see startup failures with the below exception. I tried with the latest 3.7.x, any help will be greatly appreciated.

It worked fine until 3.1.x, getting below exception anything 3.2.x beyond; I see some refactoring in the Quartz package with 3.2.x, but I need help finding the exact issue.

Can reproduce with docker mssql server container; we are on Java 17 with Hibernate.

15:47:11 ERROR {} [io.qu.ru.Application] (Quarkus Main Thread) (null) IP= reqId= Failed to start application (with profile {0}): java.lang.RuntimeException: Failed to start quarkus
        at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
        at io.quarkus.runtime.Application.start(Application.java:101)
        at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:111)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:71)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
        at io.quarkus.runner.GeneratedMain.main(Unknown Source)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at io.quarkus.runner.bootstrap.StartupActionImpl$1.run(StartupActionImpl.java:113)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.IllegalStateException: Unable to start Scheduler
        at io.quarkus.quartz.runtime.QuartzSchedulerImpl.start(QuartzSchedulerImpl.java:511)
        at io.quarkus.quartz.runtime.QuartzSchedulerImpl_Observer_start_H-6XoNLvj3wcGgHUsn1YAKHoJeM.notify(Unknown Source)
[...]

PR #1251, opened 2024-11-03
Fix #1095: 2.4 Check for MSSQLDelegate not full class name

### LINKED ISSUE #1095: SQL Server doCheckIn for update error
When quartz 2.2.3 was used to adapt sql server, this error was found in jobStoreSupport when doCheckIn(), but not in acquireNextTrigger(). Is it because org. Quartz. Impl. StdSchedulerFactory initialization time and the result of the lock handler is not set。

melloware:
Quarkus used and extended MSSQLDelegate so it skips this code:
https://github.com/quarkusio/quarkus/blob/main/extensions/quartz/runtime/src/main/java/io/quarkus/quartz/runtime/jdbc/QuarkusMSSQLDelegate.java

So we need to check for the "MSSQLDelegate" instead of the full class name. Ideally we should check for instanceof but it looks like this was intentional here using the String and before the Delegate Class is loaded

Fix https://github.com/quartz-scheduler/quartz/issues/1095
Fix https://github.com/quartz-scheduler/quartz/issues/1101

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
@@ -663,7 +663,7 @@ public abstract class JobStoreSupport implements JobStore, Constants {
             }
             
             if (getUseDBLocks()) {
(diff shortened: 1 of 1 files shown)


jhouserizer (approved):
Thanks

PR #1252, opened 2024-11-04
Fix tomorrowAt returning wrong date when called before DST change

konopka:
## Changes
- made tests repeatable by injecting fixed Clock, instead of using current time
- adjusted some javadocs @param order to match method parameter order
- changed year validation to match javadoc i.e min value 1970 and set to max value to max that is supported by java

-----------------
## Checklist
- [x ] tested locally
- [ ] updated the docs
- [x] added appropriate test
-----------------


jhouserizer (approved):
Thank you

Issue #1245, opened 2024-10-30
TriggerListener#triggerMisfired() isn't notified if a trigger misfires due to `@DisallowConcurrentExecution`


mkouba:
Steps to reproduce: 
- add a listener to the [MisfireExample](https://github.com/quartz-scheduler/quartz/blob/main/examples/src/main/java/org/quartz/examples/example5/MisfireExample.java)
- `TriggerListener#triggerMisfired()` is never fired although the triggers should fire every 3 seconds and the execution delay is `10000L`

PR #1260, opened 2024-11-10
Handle exceptions thrown while retrieving the misfired triggers

### LINKED ISSUE #1152: MisfireHandler cannot recover any triggers if one misfired trigger has errors
This is similar to #48.

In JobStoreSupport.java, `recoverMisfiredJobs` method has a loop:

```
        for (TriggerKey triggerKey: misfiredTriggers) {
            
            OperableTrigger trig = 
                retrieveTrigger(conn, triggerKey);

            if (trig == null) {
                continue;
            }

            doUpdateOfMisfiredTrigger(conn, trig, false, STATE_WAITING, recovering);

            if(trig.getNextFireTime() != null && trig.getNextFireTime().getTime() < earliestNewTime)
                earliestNewTime = trig.getNextFireTime().getTime();
        }
```

It calls `retrieveTrigger` and might throw an exception due to an invalid cron expression, for example.

Then, the entire transaction will be rolled back as there is no error/exception handling in the loop.

It will result in a defunct misfire recovery. It will run into an infinite state that tries to recover the same misfired triggers without success. All other valid triggers will not be recovered and will be stuck forever.

Questions:
1. Is this a known issue and by design?
2. If so, could you please share the background and the reason behind the decision?
3. Can we handle the `retrieveTrigger` exception in the loop and unblock the other valid triggers by skipping the erroneous ones?

Thank you!

ohksj77:
This PR...

Fix #1152

## Changes
- handled exceptions thrown while retrieving the misfired triggers
  - try-catch and logs refer to [quartznet/quartznet#1040](https://github.com/quartznet/quartznet/pull/1040/files) for consistency

-----------------
## Checklist
- [x] tested locally
- [ ] updated the docs
- [ ] added appropriate test
-----------------

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/JobStoreSupport.java
@@ -966,16 +966,25 @@ public abstract class JobStoreSupport implements JobStore, Constants {
         }
 
         for (TriggerKey triggerKey: misfiredTriggers) {
-            
-            OperableTrigger trig = 
-                retrieveTrigger(conn, triggerKey);
+            OperableTrigger trig;
+
+            try {
+                trig = retrieveTrigger(conn, triggerKey);
+            } catch (Exception e) {
+                getLog().error("Error retrieving the misfired trigger: {}", triggerKey, e);
+                continue;
+            }
 
             if (trig == null) {
                 continue;
             }
 
-            doUpdateOfMisfiredTrigger(conn, trig, false, STATE_WAITING, recovering);
-
+            try {
+                doUpdateOfMisfiredTrigger(conn, trig, false, STATE_WAITING, recovering);
+            } catch (Exception e) {
+                getLog().error("Error updating misfired trigger: {}", trig.getKey(), e);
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
If you change your comment from `Fixes issue #1152` to `Fix #1152` it will link this PR to the ticket.

ohksj77:
> Can you also make another PR for the 2.4 branch of this fix?

Okay, I'll make a PR for the 2.4 branch too.

melloware:
@ohksj77 this looks great!  do you mind checking this ticket and see if your fixes fix this issue: https://github.com/quartz-scheduler/quartz/issues/1245

ohksj77:
@melloware My fix doesn't seem to affect the issue.
I tried running both the version I fixed and the previous version, following the reproduction steps, and found the same results in both cases.

melloware:
Ok thanks for checking!

jhouserizer (approved):
Thanks. This is an improvement.  But there is still some strange/undefined ambiguity around what does and what should happen to triggers that fail to load during misfire handling ... someday somebody needs to think more about that. :)

PR #1291, opened 2024-12-29
[Feat] TriggerBuilder: TriggerBuilder's startAt Method Settings Type Extension

hwangjeyeon:
This PR extends the triggerBuilder's startAt type setting. Previously, only Date types could be delivered, so the user did the type conversion, but now LocalDateTime and ZonedDateTime can also be delivered without type conversion.

## Changes
= TriggerBuilder's startAt settings can be delivered to LocalDatetTime type
= TriggerBuilder's startAt settings can be delivered to ZonedDateTime type

## Checklist
- [x] tested locally
- [x] updated the docs
- [x] added appropriate test
- [x] signed-off on the DCO referenced in the CONTRIBUTING link below via git commit -s on my commits, and submit this code under terms of the Apache 2.0 license and assign copyright to the Quartz project owners

### DIFF
--- a/quartz/src/main/java/org/quartz/TriggerBuilder.java
+++ b/quartz/src/main/java/org/quartz/TriggerBuilder.java
@@ -18,6 +18,9 @@
 
 package org.quartz;
 
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.util.Date;
 
 import org.quartz.spi.MutableTrigger;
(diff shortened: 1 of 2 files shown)


mjeffrey:
I don't understand the use case for LocalDateTime. LocalDateTime has very few real world use cases and is typically used incorrectly. ZonedDateTime is also somewhat misunderstood. 
I think it would be more useful to use Instant as a parameter instead of the other two.

hwangjeyeon:
Thank you for your review.

I initially used LocalDateTime because the parameters I developed were primarily intended for users in the same time zone. 
However, this required additional conversion of LocalDateTime to Date, 
leading to unnecessary code in the service. So I requested this PR.

After revisiting the Java Time library, 
I now understand that using LocalDateTime and ZonedDateTime in Quartz is not appropriate. 
Since Quartz requires consistent and precise time calculations, I agree that Instant is a more suitable choice.

I will update the PR soon to provide only Instant as a parameter.
Thanks again for your valuable feedback!

VictorML11:
I agree, Instant is more convenient and in fact you can provide a custom Clock to your Instant, hence you can also specify a Zone if necessary.

VictorML11 (approved):

jhouserizer (approved):
Thank you.

PR #1302, opened 2025-01-20
Fix CVE-2019-13990 in quartz-2.1.x

henrikplate:
This is to fix vulnerability CVE-2019-13990 present in version 2.1.7.

The patch is identical to the fix provided for version 2.3.2 with fix commit [a1395ba](https://github.com/quartz-scheduler/quartz/commit/a1395ba118df306c7fe67c24fb0c9a95a4473140).

It has been tested through the down-ported unit test included in the same commit.

Disclaimer and background:
- It is clear that the given version is EOL, i.e. the project may not release a new version containing the patch, and the project generally advises users to migrate to later, supported versions. However, this fix could be a temporary remedy for users struggling to perform this migration (whatever the reason).
- This fix is part of [Endor Labs](https://www.endorlabs.com/)' efforts to create minimal invasive, yet viable security fixes for older OSS versions, similar to the well-known and established approach followed by various Linux distributions.
- By sharing the patch with the project community, we also hope to receive input as to whether the proposal is viable. It is clear, however, that the time of open source maintainers is limited. Hence, we appreciate any feedback given, but also understand if you cannot find the time to comment on merge requests opened for EOL versions.
-----------------
## Checklist
- [x] tested locally
- [ ] updated the docs
- [x] added appropriate test
-----------------

### DIFF
--- a/quartz/src/main/java/org/quartz/xml/XMLSchedulingDataProcessor.java
+++ b/quartz/src/main/java/org/quartz/xml/XMLSchedulingDataProcessor.java
@@ -186,6 +186,13 @@ public class XMLSchedulingDataProcessor implements ErrorHandler {
         
         docBuilderFactory.setAttribute("http://java.sun.com/xml/jaxp/properties/schemaSource", resolveSchemaSource());
         
+        docBuilderFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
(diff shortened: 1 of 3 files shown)


jhouserizer:
Thanks.  We'll accept this into the 2.1.x branch, but we do not plan any official releases on that line.

jhouserizer (approved):

jhouserizer (approved):

PR #1344, opened 2025-04-26
Feature - Bulk loaders

vibbix:
This PR creates a new "enhanced" bulk loader for getting jobs and triggers. It also allows preexisting methods to optionally opt-in to this reducing the number of round trips needed for loading a large selection of jobs and triggers.

## Changes
- extended test suite and added MySQL
- Added ability to get multiple JobDetails via a new method `List<JobDetail> getJobDetails(GroupMatcher<JobKey> matcher)`
- Created a delegate method called `loadExtendedTriggerPropertiesFromResultSet` 
- Created a new method called `getTriggersByJobAndTriggerGroup` that allows batch searching for triggers with Job And Group matchers
- added a `useEnhancedStatements` property for JDBC job store
-----------------
## Checklist
- [x] tested locally
- [x] updated the docs
- [x] added appropriate test
-----------------


melloware (approved):
Excellent improvement

VictorML11:
Excellent! I was actually searching for this ^^

My only suggestion is to avoid using ***** imports

quartz/src/main/java/org/quartz/impl/jdbcjobstore/DriverDelegate.java
```diff
@@ -18,26 +18,19 @@
...
-
-import org.quartz.Calendar;
-import org.quartz.Job;
-import org.quartz.JobDataMap;
-import org.quartz.JobDetail;
-import org.quartz.JobKey;
-import org.quartz.JobPersistenceException;
-import org.quartz.Trigger;
-import org.quartz.TriggerKey;
+import org.quartz.*;
```
jhouserizer: Please avoid usage of *

quartz/src/main/java/org/quartz/spi/JobStore.java
```diff
@@ -18,25 +18,16 @@
 
 package org.quartz.spi;
 
+import org.quartz.*;
```
jhouserizer: Please avoid gratuitous re-org of imports

vibbix:
All changes requested made.

jhouserizer (approved):
Looks great. Thank you very much.

quartz/src/main/java/org/quartz/impl/jdbcjobstore/Util.java
```diff
@@ -179,6 +181,59 @@ private static java.lang.reflect.Method getSetMethod(String name, PropertyDescri
...
     }
 
+    static boolean containsColumnNames(ResultSet rs, String... colNames) throws SQLException {
+        ResultSetMetaData rsmd = rs.getMetaData();
+        int columnCount = rsmd.getColumnCount();
+
+        for (int i = 1; i <= columnCount; i++ ) {
+            String name = rsmd.getColumnName(i);
+            for (String colName : colNames) {
+                if (colName.equals(name)) {
```
victorbr: Are you sure this should not be checked case-insensitive? For example, the Postgres driver has lowercase column names in the result set metadata. The arguments passed to this method are in uppercase.

fhueter:
Sadly I am a bit late to say that but if I am not wrong you have introduced a breaking change or at least a new method which should result in a minor version upgrade with that feature into the bugfix version Quartz 2.5.1 by adding a new method `public List<JobDetail> getJobDetails(GroupMatcher<JobKey> groupMatcher)` to the interface `org.quartz.Scheduler`.

I am using a dummy scheduler for the "junit" profile in my applications declared like this and I have to add (and override) that new method there - otherwise it won't compile any more:

```
	@Bean
	@Profile("junit")
	Scheduler scheduler() {
		return new DummyScheduler();
	}

	private static class DummyScheduler implements Scheduler {
		// Override all methods with dummy implementations returning null
		
		// not compiling as long as these lines have not been added to this dummy implementation
		@Override
		public List<JobDetail> getJobDetails(GroupMatcher<JobKey> groupMatcher) {
			return List.of();
		}
	}
```

Or is there a better approach to create a dummy scheduler implementation we did not find when creating that piece of code years ago?

quartz/src/main/java/org/quartz/impl/jdbcjobstore/Util.java
victorbr: Just to be clear, this is a regression in 2.5.1 which blocks upgrading to this version

Issue #1310, opened 2025-02-15
Upgrading to 2.4.0 fails due to a breaking change with com.mchange.c3p0 dependency


mossini-smeup:
Upgrading to version 2.4.0 introduces a breaking change due to the c3p0 dependency being set to provided scope. As a result, applications that relied on com.mchange.v2.c3p0.ComboPooledDataSource without explicitly declaring the dependency now fail with the following runtime error:

`java.lang.ClassNotFoundException: com.mchange.v2.c3p0.ComboPooledDataSource`

A minor version upgrade should not introduce breaking changes. If a dependency scope change leads to runtime failures, it should warrant a major version bump.

The change in dependency scope causes applications to fail unless they explicitly declare c3p0 as a dependency. This unexpected change breaks backward compatibility.

xjokerabab:
The following c3p0 package can be added to temporarily solve the problem

> 
        <dependency>
            <groupId>com.mchange</groupId>
            <artifactId>c3p0</artifactId>
            <version>0.9.5.5</version>
        </dependency>

jhouserizer:
relates to #1345

jhouserizer:
Context: .... this was a purposeful change because people who do not use DB-backed quartz (e.g. use RAMJobStore) complain about the dependency being pulled in.  Needs better documented.

PR #1345, opened 2025-05-07
Update dependency scope for DB backends

nickrobison:
When upgrading a project from Quartz 2.3.2 to 2.5.0 we ran into the following error:

```
Exception in thread "main" java.lang.NoClassDefFoundError: com/mchange/v2/c3p0/ComboPooledDataSource
	at org.quartz.utils.C3p0PoolingConnectionProvider.initialize(C3p0PoolingConnectionProvider.java:187)
	at org.quartz.utils.C3p0PoolingConnectionProvider.<init>(C3p0PoolingConnectionProvider.java:106)
	at com.rms.mapping.system.scheduler.QuartzSchedulerManager$.init(QuartzSchedulerManager.scala:225)
	at com.rms.mapping.system.scheduler.QuartzSchedulerManager$.apply(QuartzSchedulerManager.scala:201)
```

Looking at the Gradle build, it look like the `c3p0` dependency is declared as `implementation` which means it's not exposed on the runtime class path. While the workaround from #1310 is sufficient if these types are exposed in the library API then it seems to make sense to change their declaration scope.

 This PR updates the dependency declaration to restore the behavior of the 2.3.x library.

If these dependencies were intentionally hidden then perhaps a comment in the release notes or the README would be helpful?

This PR...

Fixes issue #

## Changes
-

-----------------
## Checklist
- [ ] tested locally
- [ ] updated the docs
- [ ] added appropriate test
- [ ] signed-off on the DCO referenced in the CONTRIBUTING link below via `git commit -s` on my commits, and submit this code under terms of the Apache 2.0 license and assign copyright to the Quartz project owners
-----------------

### DIFF
--- a/quartz/build.gradle
+++ b/quartz/build.gradle
@@ -19,8 +19,8 @@ dependencies {
 
     implementation "org.slf4j:slf4j-api:$slf4jVersion"
     runtimeOnly "org.slf4j:slf4j-log4j12:$slf4jVersion"
-    implementation "com.mchange:c3p0:$c3p0Version"
-    implementation "com.zaxxer:HikariCP:$hikaricpVersion"
+    api "com.mchange:c3p0:$c3p0Version"
+    api "com.zaxxer:HikariCP:$hikaricpVersion"
     compileOnly "jboss:jboss-common:$jbossVersion"
     compileOnly "jboss:jboss-minimal:$jbossVersion"
     compileOnly "jboss:jboss-system:$jbossVersion"


jhouserizer:
It was intentional - but with some hesitation.  In light of your report I will review/ponder.

jhouserizer:
The concerns here are:
* More than half of the users of Quartz use JDBCJobStore and hence end up needing these dependencies.  So its arguably a regression not to.
* A large portion (but a bit less than half) of Quartz users are using RAMJobStore and don't want these dependencies dragged in.  So it's an improvement not to.

jhouserizer:
Also relates to #1384

jhouserizer:
Hi after consideration, and discussion with a couple of others, I don't believe this change can / should be made in a dot release.  Especially considering there is another set of users that would be annoyed in the opposite direction.   Also if we were to take care of the one you mentioned, there are several others that should be reviewed.

Will prepare some comments to add to the release notes and/or installation docs to help users out.  Thanks for reporting.

jhouserizer:
Deferring for now.

PR #1389, opened 2025-09-26
fix: job is identical and replaceExisting is true

mengnankkkk:
This PR...

Fixes issue #

## Changes
-

-----------------
## Checklist
- [ ] tested locally
- [ ] updated the docs
- [ ] added appropriate test
- [ ] signed-off on the DCO referenced in the CONTRIBUTING link below via `git commit -s` on my commits, and submit this code under terms of the Apache 2.0 license and assign copyright to the Quartz project owners
-----------------


jwetzell:
Why are there 10 commits to accomplish this? And the claude settings file does not seem related so probably should not be included in this PR.

PR #1417, opened 2025-11-17
PR: Introduce pluggable TimeBroker as replacement for system clock

### LINKED ISSUE #1372: provide TimeProvider
Can we provide an interface called TimeProvider to replace the default system clock source

ThanosTsiamis:
This PR addresses Provide TimeProvider #1372, which requested a way to replace Quartz’s hard-wired use of System.currentTimeMillis() with a pluggable, testable, and customizable time source.

Quartz previously had an SPI called TimeBroker, but it was deprecated and never wired through the scheduler. This PR revives and integrates that interface as the official mechanism for controlling Quartz’s notion of “current time.”

## Important note: No change in default behavior

If no custom TimeBroker is provided, Quartz continues to use:

`SimpleTimeBroker → new Date() → System.currentTimeMillis()`

so backward compatibility is preserved.

Solves issue: #1372 

## Benefits
	•	Enables deterministic time-control in unit & integration tests
	•	Allows virtual clocks, offset clocks, or cluster-synchronized clocks
	•	Fully backward compatible
	•	No runtime or API breaking changes
	•	Resurrects an existing SPI rather than introducing a new type

-----------------
## Checklist
- [X] tested locally
- [X] updated the docs
- [X] added appropriate test
-----------------


ThanosTsiamis:
Hi there reviewers!
Please let me know if you have any comments on this (either on this PR or on the underlying issue). 
It would solve a major obstacle on a multi service project that uses quartz and needs time travel

ThanosTsiamis:
Hey maintainers! :) Any thoughts on this?
@jhouserizer @melloware

ThanosTsiamis:
## A few more words on the PR
In complex, distributed multi-service architectures, testing time-dependent logic is notoriously painful. Because Quartz has historically been hard-wired to System.currentTimeMillis(), simulating "time travel" or testing edge cases (like daylight saving shifts or specific execution gaps) requires altering the host system clock—which messes up everything else.

Deterministic Testing: A pluggable TimeBroker allows us to inject a frozen or fast-forwarded virtual clock during integration tests.

Cluster Synchronization: It gives teams the flexibility to hook into external time synchronization mechanisms if their environment demands it.

## Backwards Compatibility Check
Looking through the implementation details and the files changed, this PR is incredibly safe and fully backwards compatible:

Zero Behavior Drift by Default: If the new property `org.quartz.scheduler.timeBroker.class` isn't specified, `StdSchedulerFactory` falls back to `SimpleTimeBroker`, which relies on standard system time.

Smart Asset Reuse: Instead of introducing a brand-new, breaking API type, this simply wires up and leverages the pre-existing TimeBroker SPI.

This is a massive quality-of-life win for modern testing setups with zero risk to legacy configurations.

@jhouserizer @melloware Let me know your thoughts =]

ThanosTsiamis:
Greetings @melloware!
It has been ~300 days since the PR has been opened and no positive nor negative feedback has been made from any maintainer/developer!
Please, let me know your thoughts! =]

melloware:
@ThanosTsiamis only @jhouserizer can merge PRs. I only have review privileges.

ThanosTsiamis:
> @ThanosTsiamis only @jhouserizer can merge PRs. I only have review privileges.


Thanks for clarifying, @melloware! 🙌 =]

Since the branch requires 2 approving reviews before it can be merged, a review or feedback from your side would still be immensely helpful in moving this forward.

Whenever you have a moment, I'd really appreciate your thoughts on the code changes! If you're unavailable, is there anyone else you'd suggest reaching out to alongside @jhouserizer?

melloware (approved):

ThanosTsiamis:
@jhouserizer kind reminder for this! =]
I'd appreciate your thoughts for these changes

ThanosTsiamis:
@jhouserizer hope you're doing fine. Take a look at this PR. It got a first approval =]

ThanosTsiamis:
@jhouserizer kind reminder for this! ☝️ 
Cheers! =]

ThanosTsiamis:
@jhouserizer and whoever else can approve this=]  kind reminder for this

jhouserizer:
I made time to review it this weekend, thanks.

jhouserizer:
Hi @ThanosTsiamis ,

So, like the other times I started reviewing this, I fairly quickly run into uncertainty about which way to take this, as I quickly uncover a large number of comments, desires for changes, and even questions of value.

The TimeBroker was abandoned years ago as being impractical / unable to control all aspects of the scheduler with (more on that below), and thus other testing approaches (such as testing trigger correctness via getTimeAfter() etc.) were used (sufficiently if not awesomely) for testing quartz itself "from the inside".     It is however understood that applications/services using quartz may want a way to test behavior of the overall setup/solution they have made, and that therefore being able to (for example) advance time could be useful to that.

With that context, comments on this PR:

1- With this PR, generally, the TimeBroker is *not* controlling all aspects of the clock.  Instead it is controlling some key points where the clock is looked at, but not everything that clock has effect on. Thus, for instance it has no affect on Thread.sleep() calls that exist within the codebase, so, for example if the TimeBroker moves forward in time, those sleeps are unaffected.

2- The JavaSo, like the other times I started reviewing this, I fairly quickly run into uncertainty about which way to take this.

The TimeBroker was abandoned years ago as being impractical / unable to control all aspects of the scheduler with, and other testing approaches (such as testing trigger correctness via getTimeAfter() etc.) met (sufficiently if not awesomely) need for testing quartz itself "from the inside".     It is however understood that applications/services using quartz may want a way to test behavior of the overall setup/solution they have made, and that therefore being able to (for example) advance time could be useful to that.

With that context, comments on this PR:

1- Generally, the TimeBroker is *not* controlling all aspects of the clock.  For instance it has no affect on many Thread.sleep() calls and wait(timeOut) calls that exist within the codebase.   So if the TimeBroker moves forward in time, those sleeps are unaffected, for example.

2- Several places in the codebase also use new Date() , so they will be unaffected by the TimeBroker with this PR.  I quickly found a few examples of that, such as in Misfire handling.  Then searched the code base for "new Date" and found more than 80 instances (not including tests)

(these and other complications are why TimeBroker isn't utilized up to now)

Furthermore, there are still many cases of System.currentTimeMillis() in the codebase after this PR (such as in JobStoreSupport and RAMJobStore).

3- The TimeBroker in this PR affects what time the Scheduler thinks it is, but (because if the gap pointed out above) it doesn't affect the Scheduler's behavior it "time moves forward" etc.   To close the gaps, things would need to get notably more intertwined, with a way for the TimeBroker to signal the SchedulerThread to wake up early, etc.

4- With this PR, when QuartzSchedulerThread.currentTimeMillis() catches SchedulerException from TimeBroker it falls back to System.currentTimeMillis().   This will cause very weird behavior if it happens "mid-flight" (e.g. in the middle of testing, etc.).   Maybe something more dramatic should be done, since this would probably make everything from that point on be nonsensical.

5- The changes to QuartzScheduler.java lines 73-77 and several of the other files are superfluous and move the code to using the anti-pattern of wildcard imports.

6- The TimeBroker instance variable added by this PR to QuartzScheduler is redundant (never used) and is also superfluously final .   

7- The PR removes "@deprecated" from the TimeBroker, but javadoc at TimeBroker.java:27 still says "NOTE: TimeBroker is not currently used in the Quartz code base.


All of that said, back to the question of value (that I mentioned at the beginning of this comment).

It seems that even the changes of this PR might provide value for certain types of testing at the level of applications that use Quartz,  even though it could (with all the the above limitations) be misleading and cause users to have unfortunate expectations that cannot be met.

If you can outline some testing scenarios that these changes work for, and can make clear documentation about what works and doesn't, what the limitations are, and how to work within them,   then I'd consider merging a PR with that.


Oh, one last point (I guess really still under the category of my points 1 and 2 above), the description of the PR claims it enables "cluster-synchronized clocks" but that is misleading: a JDBC-cluster deployment with a custom TimeBroker will not get synchronized firing or check-in semantics; it gets a scheduler thread that acquires on one clock while the cluster state machine runs on another.

ThanosTsiamis:
Hi @jhouserizer  thank you for taking the time to review this and for the detailed feedback! 

Your concerns were valid, so I have reworked the PR and narrowed the contract of TimeBroker.


TimeBroker is now explicitly a scheduling clock, not a replacement for every use of time inside Quartz.
Let me first provide a simple example of the intended use case:
```
An online shop has a test environment whose logical date is January 10. 
A notification is scheduled for January 11. To test overdue-notification behavior, the application puts all Quartz nodes in standby, advances their shared logical clock to January 12, and resumes the nodes. 
Quartz then evaluates the notification using January 12 and applies its configured misfire policy, without changing the host clock.
```

The supported workflow for a discontinuous forward jump is:
1.  Put every scheduler node in standby.
2.  Apply the same logical time or offset to every node.
3.  Resume the scheduler nodes.
Resuming wakes the scheduler thread, so this PR deliberately does not attempt to support changing the clock arbitrarily while Quartz is running.


However, your comments were rather valid(thanks for that), so I made the following changes:
- RAMJobStore and JobStoreSupport now use the configured scheduling time for misfire detection.
- The built-in Simple, Cron, CalendarInterval, and DailyTimeInterval triggers use the supplied scheduling time for misfire calculations.
- Calendar replacement calculations also use the scheduling time.
- Trigger acquisition, batch windows, and job-visible fire timestamps use the scheduling clock.
- A TimeBroker failure no longer silently falls back to System.currentTimeMillis(), because mixing the two timelines could produce invalid scheduling behavior.
- A null result from a broker is treated as a failure.
- The redundant scheduler field and wildcard imports were removed.
- The stale TimeBroker Javadocs were corrected.
- The claim that this provides cluster clock synchronization was removed.
- Documentation now describes the supported scenario, required coordination, and limitations.


I also audited the roughly 80 remaining occurrences of new Date() and System.currentTimeMillis(). They should not all be replaced mechanically because many of them do not represent Quartz scheduling decisions:

- Trigger builders, XML/JMX entry points, and public trigger-calculation methods provide client-side defaults before a scheduler instance and thuss a scheduler-specific broker is available.
- Cluster check-ins, retry delays, waits, runtime measurement, database coordination timestamps, IDs, metrics, and notifications require real elapsed or wall-clock time.
- The original SPI methods continue to use wall time when called directly, preserving their existing behavior. Quartz’s internal paths call new broker-aware overloads with an explicit scheduling time.
Replacing those operational timestamps would incorrectly make cluster health, timeout, and retry behavior follow virtual time. This revision intentionally avoids that. The occurrences involved in scheduling decisions. Trigger acquisition, misfire detection and calculation, calendar updates, batch windows, and job-visible fire timestamps now use TimeBroker.


Note that in particular, this does not change JDBC cluster check-in or failure detection semantics. A clustered application remains responsible for ensuring that every node observes the same scheduling time before the nodes are resumed. Standby also does not suspend every JDBC maintenance thread, so updating the nodes still needs to be coordinated externally.

Compatibility is preserved:
- The default SimpleTimeBroker retains the existing system clock behavior.
- The existing SchedulerSignalerImpl constructor remains available and is not deprecated.
- The new SchedulerSignaler and OperableTrigger methods are default methods, so existing implementations continue to work.
- No database schema or serialized trigger state changed.
- Custom brokers must be thread--safe because scheduler and job-store threads may query them concurrently.
I added some tests covering:
- frozen scheduling time,  standby -> advance - > resume ,  RAM and JDBC misfire handling , simple and Cron misfire calculations and  calendar updates using the scheduling clock


_Again_, the intended value is therefore deliberately limited: application level tests can advance Quartz’s scheduling timeline without changing the operating-system clock or rewriting every stored trigger. It does not attempt to virtualize every timing mechanism inside Quartz. That's not what we wanted in the first place after all.

ThanosTsiamis:
@jhouserizer let me also know whether there's a better means of communication you'd prefer instead of GitHub pr comments

PR #1435, opened 2025-12-09
Limit result set size while fetching misfires jobs

kkalin68:
This PR limits the result dataset while fetching misfire jobs from the table by requested amount. 

Fixes issue #**NA**

## Changes
- Considering that `maxMisfiresToHandleAtATime` is 20 JDBC driver will load more records into the buffer than the actual limit and consumes more memory than it is really needed. 
- Another performance aspect is the query causes `filesort` due to `PRIORITY DESC` on MySQL which forces reads all eligible records from the table. The query runs under `TRIGGER_ACCESS` lock which affects the overall the library performance to process the jobs. By setting the limit MySQL will read only N records before doing `filesort` which reduces the query time significantly

-----------------
## Checklist
- [X] tested locally as well as under performance test (more 1 million of misfired jobs)
- [ ] updated the docs - **NA**
- [ ] added appropriate test - No tests are required
-----------------


kkalin68:
Also I have more changes that significantly improves Quartz performance even under unlucky condition with a lot of misfired jobs. But they are more involved and require the conversation about some major changes in handling Job priority and usage of `TRIGGER_ACCESS` lock by Worker threads.

melloware (approved):

PR #1436, opened 2025-12-10
Optimize removeTrigger method

kkalin68:
This PR optimizes `removeTrigger` method.

Fixes issue # **NA**

## Changes
- `removeTrigger` method checks if there are more tiggers associated with a job every time when a trigger is being removed. Although `SELECT COUNT` is technically correct it will be always slower and need more IO versus the proposed change. The change reduces amount of IO which is important on Cloud based DB (aka RDS) as well as improve the performance if a job has a few thousands of triggers 

-----------------
## Checklist
- [X] tested locally as well as under performance test using a real scenario
- [ ] updated the docs - NA
- [X] added appropriate test - the change is covered by existing Unit and Integration tests
-----------------

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/DriverDelegate.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/DriverDelegate.java
@@ -607,6 +607,20 @@ public interface DriverDelegate {
 
     /**
      * <p>
+     * Check if there are more tiggers associated with a given job.
+     * </p>
+     *
+     * @param connection
+     * @param jobKey
+     * @return
+     * @throws SQLException
+     */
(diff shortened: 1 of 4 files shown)


melloware:
@kkalin68 i have never heard select count() was slower as most DB's like RDS are super optimized for this especially if selecting by an indexed field?

I see you are basically doing a "select TOP 1" query so it finds something immediately and exits.

melloware (approved):

kkalin68:
> @kkalin68 i have never heard select count() was slower as most DB's like RDS are super optimized for this especially if selecting by an indexed field?
> 
> I see you are basically doing a "select TOP 1" query so it finds something immediately and exits.

It is not slower if you need to know the count. Here the count was used to see if any records exist and `select count` is overkill since it will do full index scan of all eligible records.

melloware:
absolutely i like it.

PR #1465, opened 2026-02-13
fix #1463, BLOB trigger reads from wrong ResultSet

### LINKED ISSUE #1457: **"Bug: Incorrect ResultSet used in handleTrigger() causes 'Column BLOB_DATA not found' error"**
### Environment
- **Quartz Version:** 2.5.2
- **Database:** MySQL 8.0+
- **JDBC Driver:** MySQL Connector/J
- **Java Version:** 21
- **Spring Boot:** 3.5.8

### Description
When the Quartz scheduler attempts to retrieve custom BLOB triggers (triggers that extend `AbstractTrigger`), a `SQLException` is thrown with the message "Column 'BLOB_DATA' not found".

The error occurs in `StdJDBCDelegate.handleTrigger()` method at line 1933, where the wrong ResultSet object is used to read the BLOB data.

### Root Cause
In `StdJDBCDelegate.java`, method `handleTrigger()`, lines 1927-1935:

```java
if (triggerType.equals(TTYPE_BLOB)) {
    try (PreparedStatement pps = conn.prepareStatement(rtp(SELECT_BLOB_TRIGGER))) {
        pps.setString(1, triggerKey.getName());
        pps.setString(2, triggerKey.getGroup());
        ResultSet rss = pps.executeQuery();

        if (rss.next()) {
            trigger = (OperableTrigger) getObjectFromBlob(rs, COL_BLOB);  // ❌ BUG: using 'rs' instead of 'rss'
        }
    }
}

### LINKED ISSUE #1463: BLOB Trigger does not work as it tries to read the column from the wrong result set
In the method:

handleTrigger(ResultSet rs, Connection conn, TriggerKey triggerKey, PreparedStatement ps);

```
    if (triggerType.equals(TTYPE_BLOB)) {
            try (PreparedStatement pps = conn.prepareStatement(rtp(SELECT_BLOB_TRIGGER))) {
                pps.setString(1, triggerKey.getName());
                pps.setString(2, triggerKey.getGroup());
                ResultSet rss = pps.executeQuery();

                if (rss.next()) {
                    trigger = (OperableTrigger) getObjectFromBlob(rs, COL_BLOB);
                }
            }
        }
```
The following line

`trigger = (OperableTrigger) getObjectFromBlob(rs, COL_BLOB);`

should be:

`trigger = (OperableTrigger) getObjectFromBlob(rss, COL_BLOB);`

as the ResultSet rs is the one that is passed to it, not the one it just executed.

This then throws an exception: Invalid column name,

### LINKED ISSUE #1483: [Support] quartz release
Hello, may I ask when Quartz will release a new version? Could you please release a new version because the issue at https://github.com/quartz-scheduler/quartz/issues/1457 affects the basic functionality of scheduled tasks.

e210:
This PR fixes BLOB-type triggers failing because `getObjectFromBlob` was called on the wrong `ResultSet`.

Fix #1463
Fix #1457

  ## Changes
  - Use the correct `ResultSet` (`rss`) instead of the outer one (`rs`) when reading blob data in `StdJDBCDelegate.selectTrigger`
  - Added unit test to verify `getObjectFromBlob` is called on the blob trigger's own `ResultSet`

  -----------------
  ## Checklist
  - [x] tested locally
  - [ ] updated the docs
  - [x] added appropriate test
  - [x] signed-off on the DCO referenced in the CONTRIBUTING link below via `git commit -s` on my commits, and submit this code under terms of the
  Apache 2.0 license and assign copyright to the Quartz project owners
  -----------------

### DIFF
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/StdJDBCDelegate.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/StdJDBCDelegate.java
@@ -1930,7 +1930,7 @@ public class StdJDBCDelegate implements DriverDelegate, StdJDBCConstants {
                 ResultSet rss = pps.executeQuery();
 
                 if (rss.next()) {
-                    trigger = (OperableTrigger) getObjectFromBlob(rs, COL_BLOB);
+                    trigger = (OperableTrigger) getObjectFromBlob(rss, COL_BLOB);
                 }
             }
         } else {
--- a/quartz/src/test/java/org/quartz/impl/jdbcjobstore/StdJDBCDelegateTest.java
+++ b/quartz/src/test/java/org/quartz/impl/jdbcjobstore/StdJDBCDelegateTest.java
@@ -92,6 +92,56 @@ public class StdJDBCDelegateTest  {
         assertNull(trigger);
 
     }
+
+    @Test
+    void testHandleTriggerBlob() throws SQLException, JobPersistenceException, IOException, ClassNotFoundException {
+        StdJDBCDelegate jdbcDelegate = new StdJDBCDelegate();
+        jdbcDelegate.initialize(LoggerFactory.getLogger(getClass()), "QRTZ_", "TESTSCHED", "INSTANCE", new SimpleClassLoadHelper(), false, "");
+
+        Connection conn = mock(Connection.class);
+        PreparedStatement ps = mock(PreparedStatement.class);
+        ResultSet rs = mock(ResultSet.class);
+        
+        PreparedStatement pps = mock(PreparedStatement.class);
+        ResultSet rss = mock(ResultSet.class);
+
+        // Mock basic trigger data in rs
+        when(rs.getString(Constants.COL_JOB_NAME)).thenReturn("jobName");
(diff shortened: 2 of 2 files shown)


melloware (approved):

hiSandog:
The code fix is nicely targeted. The new test would be clearer if the comments were updated to describe the post-fix behavior; right now a few lines still say the bug path is calling `getObjectFromBlob(rs, COL_BLOB)`, which is no longer what the test expects.

It may also be worth keeping the explicit `verify(rs, never()).getBlob(Constants.COL_BLOB)` assertion instead of leaving it commented out. That would make the regression check directly express the important contract: BLOB data must be read from `rss`, not the outer trigger result set.

melloware:
@e210 can you do a push on this PR it looks like the build failed but it might have been a hiccup?

e210:
Done — rebased onto latest `main` (`9294eac4`) and force-pushed. All checks are green now (Java 11/17/21 + DCO), so it looks like it was indeed a CI hiccup. Thanks for the review!

jhouserizer (approved):

PR #1470, opened 2026-02-17
fix #1349, JobInterruptMonitorPlugin memory leak and concurrency issues

### LINKED ISSUE #1349: JobInterruptMonitorPlugin can cause memory leaks
`JobInterruptMonitorPlugin` can cause memory leaks due to the fact stated in java docs for [`ScheduledThreadPoolExecutor` ](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ScheduledThreadPoolExecutor.html)

> When a submitted task is cancelled before it is run, execution is suppressed. By default, such a cancelled task is not automatically removed from the work queue until its delay elapses. While this enables further inspection and monitoring, it may also cause unbounded retention of cancelled tasks. To avoid this, use [setRemoveOnCancelPolicy(boolean)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ScheduledThreadPoolExecutor.html#setRemoveOnCancelPolicy(boolean)) to cause tasks to be immediately removed from the work queue at time of cancellation.

I believe it should be initialised this way:
```java
 this.executor = Executors.newScheduledThreadPool(1);
 ((ScheduledThreadPoolExecutor) this.executor).setRemoveOnCancelPolicy(true);
```

Other thing I've noticed that it does use 
```java
context.getJobDetail().getJobDataMap()
```
instead of recommended 
```java
context.getMergedJobDataMap()
```

e210:
This PR fixes three bugs in `JobInterruptMonitorPlugin` as reported in #1349.

  Fixes issue #1349

  ## Changes
  - Set `setRemoveOnCancelPolicy(true)` on `ScheduledThreadPoolExecutor` so cancelled tasks are immediately removed from the work queue, preventing memory leaks
  - Replaced the single `ScheduledFuture` field with a `ConcurrentHashMap<String, ScheduledFuture>` keyed by `fireInstanceId`, so concurrent jobs no longer overwrite each
   other's monitor
  - Changed `getJobDetail().getJobDataMap()` to `getMergedJobDataMap()` so that `AutoInterruptable` and `MaxRunTime` can be configured on the trigger as well as the job
  - Fixed debug log to report the actual delay used instead of always logging the default value

  -----------------
  ## Checklist
  - [x] tested locally
  - [ ] updated the docs
  - [x] added appropriate test
  - [x] signed-off on the DCO referenced in the CONTRIBUTING link below via `git commit -s` on my commits, and submit this code under terms of the Apache 2.0 license and
  assign copyright to the Quartz project owners
  -----------------

### DIFF
--- a/quartz/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
+++ b/quartz/src/main/java/org/quartz/plugins/interrupt/JobInterruptMonitorPlugin.java
@@ -16,9 +16,11 @@
  */
 package org.quartz.plugins.interrupt;
 
-import java.util.concurrent.Executors;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 import org.quartz.JobExecutionContext;
@@ -54,7 +56,7 @@ public class JobInterruptMonitorPlugin extends TriggerListenerSupport implements
     private ScheduledExecutorService executor;
 
     @SuppressWarnings("rawtypes")
-    private ScheduledFuture future;
+    private final ConcurrentMap<String, ScheduledFuture> futures = new ConcurrentHashMap<>();
 
     private Scheduler scheduler;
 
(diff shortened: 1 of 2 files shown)


melloware (approved):
Thank you for the PR

e210:
Rebased onto latest `main` (`9294eac4`) and force-pushed — the diff is unchanged from the version you approved, only the base commit moved.

The earlier red build wasn't caused by this PR: every build in that window failed (#1455, #1466–#1469 as well) and the pipeline recovered on its own by Feb 23. All checks are green now.

@melloware this one has been sitting since February — anything else needed from my side?

melloware:
No it looks good. I don't have commit permissions only review and tagging permissions. Only @jhouserizer can merge PRs

PR #1494, opened 2026-07-10
Harden JDBCJobStore deserialization: centralize getObjectFromBlob and support a JEP-290 ObjectInputFilter

### LINKED ISSUE #1495: Harden JDBCJobStore deserialization: centralize getObjectFromBlob and support a JEP-290 ObjectInputFilter
## Summary

`JDBCJobStore` deserializes the job-data-map, trigger and calendar BLOBs it reads back from the database with a raw `new ObjectInputStream(...).readObject()`, with no look-ahead deserialization filter. The same unfiltered pattern is duplicated across the base delegate and every database-specific delegate, so there is currently no single place where a filter could be enforced (CWE-502 / JEP-290 gap).

The read sites (all on the current `main`):

- `StdJDBCDelegate.getObjectFromBlob()`, the base implementation, reached for `COL_JOB_DATAMAP` (selectJobDetail), `COL_BLOB` (selectTrigger) and `COL_CALENDAR` (selectCalendar).
- The per-dialect overrides that duplicate the same `new ObjectInputStream(binaryInput).readObject()`: `PostgreSQLDelegate`, `MSSQLDelegate`, `OracleDelegate`, `HSQLDBDelegate`, `SybaseDelegate`, `WebLogicDelegate`, `PointbaseDelegate`, `CUBRIDDelegate`, `CacheDelegate`, `GaussDBDelegate` (the last added recently via #1355, i.e. new delegates keep copying the pattern).

## Why

Java object deserialization of data that may be influenced by an attacker is the classic gadget-chain risk. For the JDBC job store this is defense-in-depth (it presumes influence over the stored BLOBs, e.g. via a shared/compromised job-store database), but there is currently no way to constrain it at all short of a JVM-wide `jdk.serialFilter`. Since Java 9 (`ObjectInputFilter`, this project targets Java 11) a per-store filter is a standard, low-cost hardening, and having one central choke point is the prerequisite for it.

## Proposal

1. Centralize the deserialization into a single protected helper, `StdJDBCDelegate.readObjectFromBinaryStream(InputStream)`, that every delegate calls instead of duplicating `new ObjectInputStream(...).readObject()`. The dialect delegates keep their own dialect-specific stream-fetching; only the `readObject` step is shared.
2. Support an optional `ObjectInputFilter` on the delegate, configured through the existing delegate init-string as `objectInputFilter=<JEP-290 pattern>` (e.g. `org.quartz.jobStore.driverDelegateInitString = objectInputFilter=maxdepth=20;java.**;org.quartz.**;!*`). When unset, behavior is identical to today.

This also gives a natural home for the class-loading fix proposed in #1153 (`ClassLoaderObjectInputStream`), which touched exactly these delegates but was closed only for procedural reasons (ownership change / re-signing); both concerns can live behind the one helper.

## Open question (the reason I'm opening an issue rather than just a PR)

What should the **default** be? Because a job-data map serialized with `useProperties=false` can legitimately contain arbitrary application classes, a strict default allow-list would break existing users, so the draft below defaults the filter to **null (no change)** and leaves it opt-in. Options, happy to implement whichever you prefer:

- (a) keep it opt-in (draft as-is), zero behavior change;
- (b) apply a conservative, non-breaking resource-limit default (e.g. a generous `maxdepth`/`maxrefs`) that never rejects legitimate object graphs but stops depth/reference bombs, still fully overridable;
- (c) a documented default reject-list of the well-known gadget packages.

## Draft PR

I've opened a draft PR implementing option (a): https://github.com/quartz-scheduler/quartz/pull/1494. It compiles on the Java 11 toolchain and `StdJDBCDelegateTest` passes (7/7), including two added tests proving (1) no filter = unchanged round-trip and (2) a configured filter rejects a non-allow-listed class while still deserializing legitimate data. Commits are DCO signed-off. I'll adjust the default policy to whatever the team prefers before it leaves draft.

Nexory:
## Summary

This centralizes the `JDBCJobStore` BLOB deserialization behind a single helper and adds support for a configurable JEP-290 `ObjectInputFilter`, so the job-data-map / trigger / calendar deserialization can be constrained in one place instead of being an unfiltered `new ObjectInputStream(...).readObject()` duplicated across the base delegate and every dialect delegate (CWE-502 hardening).

Discussion / design question (default policy): see the linked issue. **Opening as a draft**: the default is left as opt-in (no behavior change) pending the team's preference on the default filter policy.

## Changes

- Add `StdJDBCDelegate.readObjectFromBinaryStream(InputStream)`, the single choke point that constructs the `ObjectInputStream`, applies the configured `ObjectInputFilter` (if any), and reads the object.
- Route the base `getObjectFromBlob()` and all ten dialect overrides (`PostgreSQLDelegate`, `MSSQLDelegate`, `OracleDelegate`, `HSQLDBDelegate`, `SybaseDelegate`, `WebLogicDelegate`, `PointbaseDelegate`, `CUBRIDDelegate`, `CacheDelegate`, `GaussDBDelegate`) through it. Each delegate keeps its own dialect-specific stream fetching; only the deserialization step is shared.
- Add an optional `objectInputFilter` field, configurable via the delegate init-string setting `objectInputFilter=<JEP-290 pattern>` (parsed with `ObjectInputFilter.Config.createFilter`). The init-string split is changed to `split("=", 2)` so filter patterns (which contain `=`) survive parsing.
- No new dependencies (`java.io.ObjectInputFilter` is JDK 9+, this project targets Java 11).

## Behavior

- **Default is unchanged.** With no `objectInputFilter` configured, the field is `null` and deserialization behaves exactly as before.
- When configured (e.g. `org.quartz.jobStore.driverDelegateInitString = objectInputFilter=maxdepth=20;java.**;org.quartz.**;!*`), the filter is applied at the single choke point for every BLOB the store reads back.

## Testing

- `./gradlew :quartz:compileJava` succeeds on the Java 11 toolchain.
- `./gradlew :quartz:test --tests StdJDBCDelegateTest`: 7 passed, 0 failed. Two tests were added:
  - `testReadObjectFromBinaryStreamWithoutFilterIsBehaviorPreserving`: legitimate map round-trips unchanged with no filter.
  - `testReadObjectFromBinaryStreamAppliesConfiguredFilter`: a legitimate map still deserializes, while a class outside the configured allow-list is rejected (`InvalidClassException`).

## Notes

Commits are DCO signed-off. This is a draft pending the default-policy discussion in the linked issue; happy to adjust the default (opt-in vs. a non-breaking resource-limit default vs. a reject-list) before finalizing.

### DIFF
--- a/examples/src/main/resources/org/quartz/examples/example13/instance1.properties
+++ b/examples/src/main/resources/org/quartz/examples/example13/instance1.properties
@@ -22,6 +22,13 @@ org.quartz.jobStore.misfireThreshold: 60000
 
 org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreTX
 org.quartz.jobStore.driverDelegateClass=org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
+
+# Optional: restrict which classes JDBCJobStore will deserialize from the database
+# (job data maps, triggers, calendars) using a JEP-290 look-ahead ObjectInputFilter.
+# It is off by default; when enabled, only classes matching the pattern are allowed.
+# Tailor the allow-list to the classes your jobs actually persist, then reject the
+# rest with "!*". The value uses java.io.ObjectInputFilter pattern syntax, e.g.:
+#org.quartz.jobStore.driverDelegateInitString=objectInputFilter=java.lang.*;java.util.*;org.quartz.**;com.myapp.jobs.**;!*
 org.quartz.jobStore.useProperties=false
 org.quartz.jobStore.dataSource=myDS
 org.quartz.jobStore.tablePrefix=QRTZ_
--- a/examples/src/main/resources/org/quartz/examples/example13/instance2.properties
+++ b/examples/src/main/resources/org/quartz/examples/example13/instance2.properties
@@ -22,6 +22,13 @@ org.quartz.jobStore.misfireThreshold: 60000
 
 org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreTX
 org.quartz.jobStore.driverDelegateClass=org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
+
+# Optional: restrict which classes JDBCJobStore will deserialize from the database
+# (job data maps, triggers, calendars) using a JEP-290 look-ahead ObjectInputFilter.
+# It is off by default; when enabled, only classes matching the pattern are allowed.
+# Tailor the allow-list to the classes your jobs actually persist, then reject the
+# rest with "!*". The value uses java.io.ObjectInputFilter pattern syntax, e.g.:
+#org.quartz.jobStore.driverDelegateInitString=objectInputFilter=java.lang.*;java.util.*;org.quartz.**;com.myapp.jobs.**;!*
 org.quartz.jobStore.useProperties=false
 org.quartz.jobStore.dataSource=myDS
 org.quartz.jobStore.tablePrefix=QRTZ_
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/CUBRIDDelegate.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/CUBRIDDelegate.java
@@ -58,9 +58,7 @@ public class CUBRIDDelegate extends StdJDBCDelegate {
         if (bytes != null && bytes.length != 0) {
             binaryInput = new ByteArrayInputStream(bytes);
 
-            try (ObjectInputStream in = new ObjectInputStream(binaryInput)) {
-                obj = in.readObject();
-            }
+            obj = readObjectFromBinaryStream(binaryInput);
         }
 
         return obj;
--- a/quartz/src/main/java/org/quartz/impl/jdbcjobstore/CacheDelegate.java
+++ b/quartz/src/main/java/org/quartz/impl/jdbcjobstore/CacheDelegate.java
@@ -22,7 +22,6 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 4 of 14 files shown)


Nexory:
Marking this ready for review. There has been no feedback on the default-policy question in #1495, so I have kept the non-breaking opt-in default (option a): the filter defaults to `null`, i.e. zero behavior change, and is fully overridable via the `objectInputFilter=<pattern>` delegate setting. CI is green across Java 11/17/21 and `StdJDBCDelegateTest` passes (7/7).

Happy to switch the default to the conservative resource-limit variant (b) or a documented reject-list (c) if the team prefers; either is a small follow-up on top of this change. Whatever default you choose, the central choke point stays the same.

melloware:
LGTM but i guess this needs the opinion of the maintainers.

Nexory:
@jhouserizer when you have a moment I'd appreciate a review. The only open item is the default
filter policy (#1495); I've kept it as a zero-behavior-change opt-in (defaults to null, fully
overridable via the objectInputFilter delegate setting), so it is safe to merge as-is. If you'd
prefer a conservative resource-limit default instead, that's a small follow-up and I'm happy to
adjust to whatever you pick. Thanks!

jhouserizer:
@Nexory  - This looks good - I'm remote next few days (just on my phone )- I think how you've defaulted it is fine - would you mind adding a bit of update to doc/examples about how to use it?  Even just incorporating the example into an existing one, or something? e.g.  https://github.com/quartz-scheduler/quartz/tree/main/examples , https://github.com/quartz-scheduler/quartz-scheduler.org-site/tree/master/documentation/quartz-2.5.x/configuration

Nexory:
Thanks! Added a commented, off-by-default example of the `objectInputFilter` setting to the JDBC/clustering example (example13), showing the `ObjectInputFilter` pattern syntax on `driverDelegateInitString`. It stays disabled by default, so the example runs unchanged.

PR #1498, opened 2026-08-03
Add RFC 5545 RRULE trigger support (via org.dmfs:lib-recur)

sideeffffect:
## Summary

Adds first-class support for **RFC 5545 `RRULE`** (iCalendar recurrence rules) as a scheduling option, mirroring the existing `CronTrigger` stack. New public API:

| RRULE class | Cron analogue |
|---|---|
| `RRuleExpression` | `CronExpression` |
| `RRuleTrigger` / `RRuleTriggerImpl` | `CronTrigger` / `CronTriggerImpl` |
| `RRuleScheduleBuilder` | `CronScheduleBuilder` |
| `RRuleCalendar` | `CronCalendar` |
| `RRuleTriggerPersistenceDelegate` | `CronTriggerPersistenceDelegate` |
| `RRuleTriggerSupport` (JMX) | `CronTriggerSupport` |

```java
Trigger trigger = newTrigger()
        .withIdentity("myTrigger")
        .withSchedule(rruleSchedule("FREQ=WEEKLY;BYDAY=MO,WE,FR;BYHOUR=10;BYMINUTE=0"))
        .startAt(futureDate(10, MINUTES))
        .build();
```

## Design notes

- **Recurrence engine:** delegated to [`org.dmfs:lib-recur`](https://github.com/dmfs/lib-recur) `0.17.1`. This class adapts it to Quartz idioms and adds RFC-level validation for the parts where lib-recur is more permissive than RFC 5545 (out-of-range `BY*` values, `BYWEEKNO` / ordinal `BYDAY` on the wrong `FREQ`).
- **`DTSTART` anchor:** RFC 5545 defines an `RRULE` relative to a `DTSTART`. In Quartz that anchor is supplied by the trigger's mandatory `startTime`, exactly as a `CronTrigger` uses its start-time as a lower bound. Following the RFC, the start instant only fires if it actually matches the pattern.
- **Persistence:** a new `QRTZ_RRULE_TRIGGERS` table (RRULE string + time-zone) with DDL added to every vendor script under `jdbcjobstore`, a registered `RRuleTriggerPersistenceDelegate`, and a new `TTYPE_RRULE` discriminator. The delegate loads via its own dedicated `SELECT` (rather than the inlined bulk join) because the shared `TIME_ZONE_ID` column would collide in `SELECT_BULK_TRIGGERS_BASE`.
- Supports `FREQ`, `INTERVAL`, `COUNT`, `UNTIL`, `WKST`, `BYMONTH`, `BYWEEKNO`, `BYYEARDAY`, `BYMONTHDAY`, `BYDAY` (incl. ordinals), `BYHOUR`, `BYMINUTE`, `BYSECOND`, `BYSETPOS`.

## Tests

Unit tests for the expression, schedule builder, trigger and calendar, plus a Derby JDBC integration test exercising the persistence delegate end to end.

## Note: two alternative PRs

This is **one of two alternative PRs** implementing the same feature and public API, so the maintainers can choose an approach:

- **This PR** — uses the `org.dmfs:lib-recur` library.
- **#1499** — a dependency-free, from-scratch RFC 5545 engine.

Both expose an identical public API and pass the identical test suite; only `RRuleExpression`'s internals (and the added dependency) differ.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sideeffffect:
Closing in favor of the dependency-free implementation in #1499, which the maintainers preferred (see discussion there). Both PRs shared an identical public API and test suite; #1499 carries the same feature without adding a third-party dependency.

